PR #7, opened 2023-04-05
🚧 Deploy image to the internal docker registry

shepilov:


twake/docker/twake-node/Dockerfile
```diff
@@ -36,20 +36,6 @@ RUN NODE_ENV=development npm install && \
...
-    yarn global add webpack webpack-cli jest && \
-    cp src/app/environment/environment.ts.dist src/app/environment/environment.ts && \
-    export NODE_ENV=production && \
-    yarn build && \
-    rm -R node_modules && \
-    mv build/* ../app/public && \
-    cd .. && rm -R public_build/
-
 # Development Stage
 FROM with-frontend as development
```
RomaricMourgues: If we remove the "Add frontend into node" I guess the "FROM with-frontend as development" isn't relevant anymore.
- Replace "FROM node-base as with-frontend" by "FROM node-base as installed-libs"
- Remove lines 29 to 38
- Replace the two "FROM with-frontend" with "FROM installed-libs"

RomaricMourgues (changes requested):
I propose to also rename this folders:
- docker/twake-node to docker/tdrive-node
- docker/twake-nginx to ...

Also we can remove `/.github/workflows/integration-test.yml` which is not relevant anymore (or we need to reimport and clear the content of https://github.com/linagora/Twake/tree/main/twake/docker/integration-test this integration tests was only doing login + creation of a channel, maybe better to remove it as we'll convert the frontend to Futter)

shepilov:
> I propose to also rename this folders:
> 
> * docker/twake-node to docker/tdrive-node
> * docker/twake-nginx to ...
> 
> Also we can remove `/.github/workflows/integration-test.yml` which is not relevant anymore (or we need to reimport and clear the content of https://github.com/linagora/Twake/tree/main/twake/docker/integration-test this integration tests was only doing login + creation of a channel, maybe better to remove it as we'll convert the frontend to Futter)

Ok, sure, I was going to do it, and had the same question about the integration-test.yml, thank you for the quick review

twake/docker/twake-node/Dockerfile
shepilov: Why I should remove lines 29 to 38 if it is installing backend libs?

RomaricMourgues: Yep don't remove it, I don't know why I said that

RomaricMourgues:
It seems there is still a link somewhere to a "twake-node" folder

shepilov:
> It seems there is still a link somewhere to a "twake-node" folder

yep, looking

MontaGhanmy (approved):

RomaricMourgues (approved):

PR #20, opened 2023-04-14
🕹 Bootstrap static json applications definition

RomaricMourgues:
- Added documentation for plugins in https://github.com/linagora/TDrive/pull/20/files#diff-c1cc66db78e1d3bb4713c55d5ab2b8b3f00d694da66087ddf8542c0f7667c535
- First clean of Twake documentation to have a TDrive documentation
- Implemented static plugins from config/*.json files instead of from database
- Updated internal proxy to plugins when running with docker-compose or helm


RomaricMourgues:
Tested and ready

PR #59, opened 2023-05-17
🛠 Allow applications to manage the drive

RomaricMourgues:
- Applications can now manage the drive if they are installed in the company
- Application are responsible to test if the user have the right to edit/view the file themselves
- Test of applications in general to be done in another PR


chibenwa:
> Application are responsible to test if the user have the right to edit/view the file themselves

Isn't a decentralize right enforcement systema security risk?

RomaricMourgues:
In TDrive the market place of plugins is replaced by simple static in-configuration plugins. It uses a separated code base, but we can consider it as working inside TDrive and not like before where we could connect to Slack or Trello working outside TDrive.
- There is also no plan of adding any plugins beside OnlyOffice and internal smart connectors for the next 12 months.
- The other alternative is to make the app "act as the connected user" but it only solve partially the problem and Onlyoffice needs to do async operations (OK now it is only after up to one minute but it could need async operations) and this approach doesn't seem good.
- There is still an enforced per-company access limitation

So in my mind working on a new advanced application access restriction is not for now.

Issue #46, opened 2023-05-09
Create a "Shared with me" section


shepilov:
In this section user can find all the files which was shared by somebody with him.
Files in the company space or personal file

![Image](https://github.com/linagora/TDrive/assets/14924963/52a2dee7-c7ef-4472-baee-0ffababa2054)

PR #67, opened 2023-06-02
🚧 #46 Shared with me

shepilov:


tdrive/backend/node/config/test.json
```diff
@@ -19,5 +19,11 @@
     "amqp": {
       "urls": ["amqp://guest:guest@rabbitmq:5672"]
     }
+  },
+  "search": {
+    "type": "elasticsearch",
+    "elasticsearch": {
+      "endpoint": "https://localhost:9200"
+    }
```
RomaricMourgues, line 27: So we will always do tests with elastic search (no mongodb-search) ? The previous mode was mongo+mongo then cassandra+es, just to make sure I have no problem with this change :)

tdrive/backend/node/src/services/documents/entities/drive-file.search.ts
```diff
@@ -9,6 +9,9 @@ export default {
     added: entity.added,
     name: entity.name,
     company_id: entity.company_id,
+    access_entities: entity.access_info?.entities?.map(e => e.id),
```
RomaricMourgues: It will work because we mostly use uuids so it is unique to all types of entities. But it could be clearer to concat e.type + "_" + e.id.

tdrive/backend/node/src/services/documents/entities/drive-file.search.ts
```diff
@@ -9,6 +9,9 @@ export default {
...
+    last_modified: entity.last_modified,
```
RomaricMourgues, line 15: Make sure this is a number just in case I remember some fields were strings to store timestamp.

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -765,12 +768,37 @@ export class DocumentsService {
...
-        ...(options.creator ? { $in: [["creator", [options.creator]]] } : {}),
-        ...(options.added ? { $in: [["added", [options.added]]] } : {}),
-        $text: {
-          $search: options.search,
-        },
+        $in: [
+          ["access_entities", [context.user.id, context.company.id]],
+          ...(options.company_id ? [["company_id", [options.company_id]] as inType] : []),
+          ...(options.creator ? [["creator", [options.creator]] as inType] : []),
+          ...(options.added ? [["company_id", [options.added]] as inType] : []),
```
RomaricMourgues: `["company_id", [options.added]` -> `["added", [options.added]`

RomaricMourgues (commented):
Great to me, some comments just to make sure but I don't see any problem

RomaricMourgues (commented):
Great to me, some comments just to make sure but I don't see any problem

tdrive/backend/node/config/test.json
shepilov: No, it will be the same, params in test jobs still `SEARCH_DRIVER = mongodb`, it's just the default value to let skip this params in environment variables

tdrive/backend/node/src/services/documents/services/index.ts
shepilov: ✅ done

PR #106, opened 2023-06-22
🐛Removed noisy notification with shared link

jouliet:


shepilov:
It won't copy pwd on focusOut event, but ok, anyway we are going to change it, and it's better to copy URL together with pwd

Issue #115, opened 2023-06-23
When you try to move file or folder there is no need to show files in the tree view


shepilov:
You can move something only to folder

PR #135, opened 2023-06-28
:lipstick: Only folders in the tree view (#115)

Greemty:
Suppression of the code displaying all files in the tree view

### DIFF
--- a/tdrive/frontend/src/app/views/client/body/drive/modals/selector/index.tsx
+++ b/tdrive/frontend/src/app/views/client/body/drive/modals/selector/index.tsx
@@ -91,36 +91,6 @@ const SelectorModalContent = () => {
             </div>
           </div>
         ))}
-        {files.map(file => (
-          <div
-            key={file.id}
-            className={
-              'flex flex-row items-center border-t -mt-px px-4 py-2 cursor-pointer ' +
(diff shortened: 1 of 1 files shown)


tdrive/frontend/src/app/views/client/body/drive/modals/selector/index.tsx
```diff
@@ -91,36 +91,6 @@ const SelectorModalContent = () => {
             </div>
           </div>
         ))}
-        {files.map(file => (
```
shepilov, line 94: do not remove files at all from the widget, just make it optional, and for the "Move" action, set it to false

shepilov (changes requested):

tdrive/frontend/src/app/views/client/body/drive/modals/selector/index.tsx
```diff
@@ -91,6 +91,40 @@ const SelectorModalContent = () => {
             </div>
           </div>
         ))}
+        {false && ( //if false files wont display in the tree-view
```
shepilov: I mean it should be `Selector` component property.

```
so the use of this component should look like 
<SelectorModal showfiles=false />
```

shepilov (approved):

Issue #124, opened 2023-06-26
If the folder name is long, the layout is broken with the breadcrumbs


shepilov:
<img width="1432" alt="image" src="https://github.com/linagora/TDrive/assets/14924963/4364fd80-1188-4a91-8462-c87b5240d23f">

Issue #117, opened 2023-06-23
When she folder name is long, the whole layout is broken


shepilov:
<img width="1437" alt="image" src="https://github.com/linagora/TDrive/assets/14924963/3b8db745-6696-454b-9f6e-d9ac580d70f4">

shepilov:
duplicate

PR #137, opened 2023-06-29
💄Fix broken layout due to long folder/file name (#117 & #124)

Aymerix01:
## Description

Two issues were fixed:
The first one was when folder/file name were really long, the right layout with the 3 dots button was not on the screen. Now, if the name is too long, this button is still visible but not the whole name (in fact the name goes under the button and doesn't push it away)
Folder,file with long name

The second one was when a folder had also a long name and you entered into it, the access path could also push the right layout out of the screen. Now, if the folder name is too long, it ends with ... in the access path to reduce it.
access path


tdrive/frontend/src/app/views/client/body/drive/documents/document-row.tsx
```diff
@@ -104,10 +104,10 @@ export const DocumentRow = ({
           }
         />
       </div>
-      <div className="grow text-ellipsis whitespace-nowrap overflow-hidden">
+      <div className="grow text-ellipsis whitespace-nowrap overflow-hidden" style={{ width: '100px'}}>
```
shepilov: do not use absolute values at all please, absolute values are eligible only for min and max style,

tdrive/frontend/src/app/views/client/body/drive/documents/document-row.tsx
```diff
@@ -104,10 +104,10 @@ export const DocumentRow = ({
...
         <Base>{item.name}</Base>
       </div>
-      <div className="shrink-0 ml-4">
+      <div className="shrink-0 ml-4" style={{ width: '20px' }}>
```
shepilov: and also, could you move the styles, to the styles file, or what is even better if it's possible to use existing classes from tailwind css

tdrive/frontend/src/app/views/client/body/drive/header-path.tsx
```diff
@@ -25,6 +25,18 @@ export default ({
...
 };
 
+function checkNamelength (name: any){
+  if (typeof name !== "undefined" ){
+    if (name.length >= 30){
+      return name.substring(0,30)+" ...";
+    } else {
+      return name;
+    }
+  }else {
```
chibenwa:
```suggestion
  } else {
```

tdrive/frontend/src/app/views/client/body/drive/header-path.tsx
```diff
@@ -25,6 +25,18 @@ export default ({
   return <PathRender inTrash={inTrash || false} path={path} onClick={id => setParentId(id)} />;
 };
 
+function checkNamelength (name: any){
```
chibenwa:
```
function checkNamelength (name: any){
```

This is a bad name.

What are you "checking" for? Is checking even the best name here?

shepilov (approved):

Issue #96, opened 2023-06-20
Refine all the docker-compose files 


shepilov:
I want to be able to run everything in one command and this command should be not more than 40 symbols. I don't want to run env like 

- [ ] Only containers that you need to run TDrive backend locally
- [ ] Only containers that you need to run TDrive frontend locally locally
- [ ] Simple TDrive on premise with the only Mongo database

PR #139, opened 2023-06-30
Refine all the docker-compose files (#96)

Aymerix01:
The issue was that you should write docker-compose -f docker-compose.dev.mongo.yml up -d node --build, and then go to frontend (cd frontend) and launch the dev interface with yarn dev:start HTTP=false to be able to see something on http://localhost3000/ . Now, you only need to launch docker-compose up -d and go to http://localhost to see TDrive. 
TDrive refine docker

Everything works except OnlyOffice.
OnlyOffice error

Important to note, the only docker-compose file used is docker-compose.yml but i let the others in case someone needs them.


Aymerix01:
There may be a problem on Windows due to a bad convert line break on entrypoint.sh and self-signed.sh .
This error leads to an instant stop of the tdrive-frontend contener on docker and with a log in the container like this: exec /entrypoint.sh: no such file or directory (or /self-signed.sh: no such file or directory )
I solved it with Notepad++ (Open entrypoint.sh with notepad++ then edit, EOL conversion and switch to Unix (LF) )
Notepadd++ convert line break

tdrive/docker/onlyOffice-connector/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Base image
+FROM docker.io/onlyoffice/documentserver
```
chibenwa: Please use a fixed version if available

chibenwa (commented):
I would also add a friendly README in `tdrive/docker` either:

 - ... Pointing to the doc `run tdrive` if available
 - ... or documenting itself the way to run tdrive with docker-compose...

tdrive/docker/onlyOffice-connector/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# Base image
```
shepilov: why do you need a connector docker file here?
it should be only in it's repository, and you can use existing image

tdrive/docker/tdrive-frontend/Dockerfile
```diff
@@ -20,6 +20,7 @@ COPY docker/tdrive-frontend/nginx.conf /etc/nginx/nginx.conf
 COPY docker/tdrive-frontend/site.conf /etc/nginx/sites-available/site.template
 COPY docker/tdrive-frontend/redirect.conf /etc/nginx/sites-available/redirect
 COPY docker/tdrive-frontend/entrypoint.sh /
+COPY docker/tdrive-frontend/self-signed.sh /
```
shepilov, line 23: and this one also, why do you need it in root folder?

PR #141, opened 2023-07-03
✨ Implement the quota in the TDrive (#130)

Aymerix01:
The objective was to create a quota in TDrive, more precisely, create a maximum storage for a user. Now, the user has a quota of 1GB of storage and an extra of 1MB for the trash. This values can be modify in store.ts (tdrive\frontend\src\app\features\drive\state\store.ts) in the DriveQuotaStorage function. The UI looks like this:
Quota

As expected, if a user tries to add an extra file but he didn't have enough storage, he won't be able to add it to the drive. In this example, I have reduced the quota to be able to test it and I can't import a file because I have used all the storage:
QuotaLimitation

I have also solved the issue with the storage indicator when you import files in "my drive"

I have decided also that if someone shares a file with you, it won't increase your storage:
SharedDoc no limitations


tdrive/frontend/src/app/features/drive/hooks/use-drive-actions.tsx
```diff
@@ -47,15 +51,21 @@ export const useDriveActions = () => {
...
     async (item: Partial<DriveItem>, version: Partial<DriveItemVersion>) => {
-      let driveFile = null;
-      if (!item.company_id) item.company_id = companyId;
-      try {
-        driveFile = await DriveApiClient.create(companyId, { item, version });
-        await refresh(item.parent_id!);
-      } catch (e) {
+      const rootItem = await DriveApiClient.get(companyId, 'root');
+      const myDriveItem = await DriveApiClient.get(companyId, 'user_' + user?.id);
+      if (typeof item.size != 'undefined' && rootItem.item.size + item.size + myDriveItem.item.size <= mainQuota) {
```
chibenwa, line 56: Is the quota also checked on the backend side?

tdrive/frontend/src/app/features/drive/hooks/use-drive-actions.tsx
```diff
@@ -86,9 +96,16 @@ export const useDriveActions = () => {
...
     async (id: string, parentId: string) => {
+      const driveItem = await DriveApiClient.get(companyId, 'trash');
       try {
-        await DriveApiClient.remove(companyId, id);
-        await refresh(parentId || '');
+        const itemSize = (await DriveApiClient.get(companyId, id)).item.size
+        if (typeof itemSize != 'undefined' && driveItem.item.size + itemSize <= trashQuota) {
+          //ToasterService.error(Languages.t('Vous utilisez : ' + tstorageSize + ' octets de stockage'));
+          await DriveApiClient.remove(companyId, id);
+          await refresh(parentId || '');
```
chibenwa, line 105: As a user you should at the very leat inform me that my trash is full and promt me to confirm before deleting forever...

chibenwa (commented):
For me there is major security and UX topics to discuss regarding this PR.

I let JS code style comments to JSists...

PR #146, opened 2023-07-07
🕹 Manage root users

MontaGhanmy:


tdrive/backend/node/src/services/documents/web/controllers/documents.ts
```diff
@@ -175,6 +186,13 @@ export class DocumentsController {
           )
         : [],
     };
+
+    if(id == "root" && companyUserRole < 1) {
```
chibenwa: What does `1` mean?

(Magic constant spotted).

I recommend to extract this as a global variable and give it a name.

tdrive/frontend/src/app/views/client/body/drive/modals/manage-users/public-link-access.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const [useExpiration, setUseExpiration] = useState((props.expiration || 0) > 0);
+  const [expiration, setExpiration] = useState(props.expiration);
+  const handlePasswordBlur = () => {
+    props.onChangePassword(password);
+  };
+  
+
+  /* useEffect(() => {
+    props.onChangePassword(usePassword ? password : '');
+  }, [usePassword, password]); */
```
chibenwa: Remove commented code?

tdrive/backend/node/src/services/documents/web/controllers/documents.ts
```diff
@@ -187,7 +189,7 @@ export class DocumentsController {
         : [],
     };
 
-    if(id == "root" && companyUserRole < 1) {
+    if (id == "root" && companyUserRole < 1) {
```
chibenwa: What is the meaning of `1` here? 
That's the magic constant syndrom. I would recommand to:
 - 1. Strong type. It looks like some kind of permissions is encoded as an `int`. This could be made explicit with a type, that could furthermore encapsulate and centralize right checking logic, that then could (must?) be unit tested? 
 - 2. Extract constants to give them a name. Favor `enum`s if supported in typescript.

tdrive/backend/node/src/services/documents/web/controllers/documents.ts
```diff
@@ -275,6 +277,39 @@ export class DocumentsController {
...
+    const { id } = request.params;
+    const update = request.body;
+
+    if (!id) throw new CrudException("Missing id", 400);
+    if (id == "root") {
+      const companyUser = await globalResolver.services.companies.getCompanyUser(
+        { id: update.company_id },
+        { id: update.user_id },
+      );
+      if (companyUser) {
```
chibenwa, line 299: What happens if the guy is not a company user? 
We likely should return an HTTP error for this, no? `406` not acceptable ?

chibenwa (commented):
My 2 cents...

tdrive/backend/node/src/services/documents/web/controllers/documents.ts
```diff
@@ -189,7 +189,7 @@ export class DocumentsController {
         : [],
     };
 
-    if (id == "root" && companyUserRole < 1) {
+    if (id == "root" && companyUserRole == 1) {
```
chibenwa: Can we create constants for each value user role?

Or even better use an enum: https://www.typescriptlang.org/docs/handbook/enums.html ?

MontaGhanmy: The snippet above is outdated.

PR #150, opened 2023-07-11
🚸The first page(default) should be My Drive (#144)

Aymerix01:
The first page is now "My Drive" and not "Shared Drive"

### DIFF
--- a/tdrive/frontend/src/app/views/client/common/disk-usage.tsx
+++ b/tdrive/frontend/src/app/views/client/common/disk-usage.tsx
@@ -1,14 +1,16 @@
 import { Base, Title } from '@atoms/text';
 import { useDriveItem } from '@features/drive/hooks/use-drive-item';
 import { formatBytes } from '@features/drive/utils';
+import { useCurrentUser } from 'app/features/users/hooks/use-current-user';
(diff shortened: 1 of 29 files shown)


tdrive/frontend/src/app/views/client/common/disk-usage.tsx
```diff
@@ -1,14 +1,16 @@
...
 import Languages from "features/global/services/languages-service";
 
 export default () => {
   const { access, item } = useDriveItem('root');
   const { item: trash } = useDriveItem('trash');
+  const { user } = useCurrentUser();
   return (
     <>
-      {access !== 'read' && item?.id === 'root' && (
+      {access !== 'read' && (
```
shepilov, line 13: why you removed this condition?

Aymerix01: I removed it because the storage indicator doesn't show up otherwise :
DriveStorage

Furthermore, I don't really understand why item id should be root because it could also take into account the 'user_'+user.id

Finally, I didn't change it for item?.id === 'user_'+user.id because the storage indicator also doesn't show up.

shepilov (approved):

Issue #68, opened 2023-06-05
By clicking on folder in the search window the user should go to this folder in the main screen


shepilov:
<img width="1153" alt="image" src="https://github.com/linagora/TDrive/assets/14924963/984acb0a-9821-418e-88a4-1f832b56682b">

There is also no need for a download icon for folders.

PR #152, opened 2023-07-11
Go to folder when searched (#68)

Aymerix01:
The user is now able to go to a folder when he clicks on it in the search window.

Furthermore, the download button is no more present for folders:
Download Button for folder

### DIFF
--- a/tdrive/frontend/src/app/components/search-popup/parts/drive-item-result.tsx
+++ b/tdrive/frontend/src/app/components/search-popup/parts/drive-item-result.tsx
@@ -29,6 +29,7 @@ import { DriveCurrentFolderAtom } from '@views/client/body/drive/browser';
 import { useHistory } from 'react-router-dom';
 import RouterServices from '@features/router/services/router-service';
 import { useCurrentUser } from 'app/features/users/hooks/use-current-user';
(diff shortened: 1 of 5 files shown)


tdrive/frontend/src/app/components/search-popup/parts/drive-item-result.tsx
```diff
@@ -45,10 +45,15 @@ export default (props: { driveItem: DriveItem & { user?: UserType } }) => {
   const { open: openViewer } = useFileViewerModal();
   const { open } = useDrivePreview();
 
+  function openDoc(file: DriveItem){
+    open(file);
+    if (file.is_directory) props.onClose();
```
shepilov: looks like something strange, we are calling onClose property only if it is directory.
the first question - "Why?"

Aymerix01: It's because when you open a folder, you go in it and therefore you need to close the research window (modal) but if  you open a file, it opens another modal (the preview of the file) and the research window stays behind. This is a quick video to show it:

https://github.com/linagora/TDrive/assets/123820772/e3618c4b-01ab-48b1-9bb8-f24dea06aab8

If you want that the research window closes when you click on a file, I can remove the condition.

PR #189, opened 2023-09-11
🍄 Separate trash for the personal and shared drive

### LINKED ISSUE #179: [Story] Separate trash for the personal and shared drive
## User story summary
[Definition](#Definition)  
[UI/UX Design](#UI-UX-design)  

## Definition
We need to separate the trash into two different trashes

- So you have trash from your personal drive. It’s straightforward and nothing special, you should see everything you or somebody who has access to the file deleted. 
But the key point is that you are the `OWNER` of the file.
- For the shared drive,  only if you have access to it. 
You should be able to view all the files, deleted from the shared drive if you had access to them before they were deleted. 
When you try to empty the trash, you can delete only files with the `manage` access

## UI UX Design
In the trash page, we need to have a selector as in the breadcrumbs menu of the path. To choose the source from where the files were deleted :
 - My Drive
 - Shared Drive 

## Implementation

- [x] To the DriveItem model add type = "personal" | "shared" field
- [x] Also to the DriveItem use is_in_trash flag
- [x] Personal trash it will be all the files I'm a creator with the flag is_in_trash = true, shared trash we need to query all the files type == shared" and  is_in_trash = true
- [x] Only for company admins can view and modify the "shared" trash
- [x] Don't forget to modify "type" flag when moving one folder from shared to my drive and vice versa
- [x] Also modify file creation, depending where you are creating a file in your personal drive or in the shared one
- [x] Restore from the trash function also should be updated
- [ ] Update the current ones and wright new tests
- [x] Filter the current children and check that there is no files in the trash for the current "browse' query
- [ ] Update the search, and filter files that are in the trash, or what is better, add the flag is_in_trash to the search index, and do not return files that are in trash
- [ ] Update the logic for the widget with the size of the "Trash"

MontaGhanmy:


tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/cassandra/query-builder.ts
```diff
@@ -77,7 +77,7 @@ export function buildSelectQuery<Entity>(
     whereClause.trim().length ? "WHERE " + whereClause : ""
   } ${orderByClause.trim().length ? "ORDER BY " + orderByClause : ""}`
     .trimEnd()
-    .concat(";");
+    .concat(" ALLOW FILTERING;");
```
chibenwa: What is the actual query being executed?

Beware of `ALLOW FILTERING` on Cassandra, this is a **performance killer**, and a sign that the underlying data model needs to be re-thought.

I stay available might you need a Cassandra data model workshop.

tdrive/backend/node/test/unit/core/services/database/services/orm/connectors/cassandra/query-builder.test.ts
```diff
@@ -17,7 +17,7 @@ describe("The QueryBuilder module", () => {
       const result = buildSelectQuery<DriveFile>(DriveFile, filters, {}, { keyspace: "tdrive" });
 
       expect(result).toEqual(
-        "SELECT * FROM tdrive.drive_files WHERE company_id = comp1 AND parent_id = 'parent1';",
+        "SELECT * FROM tdrive.drive_files WHERE company_id = comp1 AND parent_id = 'parent1' ALLOW FILTERING;",
```
chibenwa: Sorry but this looks BAD to me.

Can you disclose the table structure? The requests you are performing on this table?

tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/cassandra/query-builder.ts
```diff
@@ -28,6 +29,7 @@ export function buildSelectQuery<Entity>(
         return;
       }
       if (isObject(filter) && JSON.stringify(filter).includes("ne")) {
+        allowFiltering = true;
```
chibenwa: Again have you been conducting serious performance benchmarks before conducting this change?

Due to the "log structured merge tree" architecture of the Cassandra storage, allow filtering can cause each request to scan through potentially terabytes of data.

As explained yesterday, I have big doubts that this change is compatible with anything looking to a serious production set up and would advise the team to look for a longer term solution.

MontaGhanmy: Generally yes, but in our case we're restricting the partition and using indexes to query the data. We're also restricting the use of Allow filtering to a very specific scenario. We won't end up scanning through terabytes of data.

shepilov:
@MontaGhanmy take a look please

When you click in the "Trash", the menu item is not highlighted
image

shepilov:
When you go to the folder inside the trash, breadcrumbs look like this

image

Without a folder name, and navigation is broken.

shepilov:
There is no way to restore items from trash.

shepilov:
Also we need to remove context menu from trash

image

shepilov:
And context menu for the items inside trash, 

image

Threre is not need to manage access, rename, share or copy link of the item that is in trash

PR #215, opened 2023-10-04
🐳  Added minio to docker compose dev file

MontaGhanmy:


shepilov:
@MontaGhanmy  could you please move it to another docker file, and keep dev with plain local storage and the most simple configuration for quickly start

PR #270, opened 2023-11-23
✉️ SMTP TLS default value

MontaGhanmy:


tdrive/backend/node/src/core/platform/services/email-pusher/index.ts
```diff
@@ -43,7 +43,7 @@ export default class EmailPusherClass
       const smtpConfig: SMTPClientConfigType = {
         host: this.configuration.get<string>("smtp_host", ""),
         port: this.configuration.get<number>("smtp_port", 25),
-        requireTLS: this.configuration.get<boolean>("smtp_tls", true),
+        requireTLS: this.configuration.get<boolean>("smtp_tls", false),
```
chibenwa, line 46: When does TDrive send mails?

PR #358, opened 2024-02-05
🐛Throw exception on S3 writing error

shepilov:


tdrive/frontend/src/app/features/auth/provider/oidc/oidc-auth-provider-service.ts
```diff
@@ -53,7 +53,7 @@ export default class OIDCAuthProviderService
         redirect_uri: getAsFrontUrl(OIDC_CALLBACK_URL),
         response_type: 'code',
         response_mode: 'query',
-        scope: 'openid profile email address phone offline_access',
+        scope: 'openid profile email address phone offline_access applications',
```
chibenwa, line 56: How is this related to the problem at hand?

chibenwa (commented):
Where is the unit test?

In James we can reproduce outages in tests by pausing dockers, for instance.

tdrive/frontend/src/app/features/auth/provider/oidc/oidc-auth-provider-service.ts
shepilov: @chibenwa  It's a quick fix, maybe also some comments about code style?

chibenwa: No I ask why we need to change the OIDC scope on an error about S3 handling.

This seems unrelated to me, I just wonna make sure it is not done by mistake.

PR #369, opened 2024-02-21
🐛 Fix files notifications on shared folder

### LINKED ISSUE #365: [Bug] notification that files was uploaded to the folder was send not only for the folder owner but to everybody with who the folder was shared

MontaGhanmy:


tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -344,19 +344,17 @@ export class DocumentsService {
               )
             : [];
 
-          if (sharedWith.length > 0) {
-            // Notify the user that the document has been shared with them
-            this.logger.info("Notifying users that the document has been shared with them: ", {
+          if (context.user.id !== parentItem?.creator && sharedWith.length > 0) {
+            // Notify the owner that the document has been shared with them
```
shepilov, line 373: @MontaGhanmy could you please right test on who is receiving notifications?

PR #435, opened 2024-03-12
Refactored build and publish workflow

tk-nguyen:
Need to add Docker Hub credentials for Linagora org as secrets: `DOCKERHUB_USERNAME` and `DOCKERHUB_PASSWORD`.

Tag pushes with changes will trigger publish workflow. Tags matching [semver naming pattern](https://semver.org/#spec-item-2) `*.*.*` will update major, minor and full version. Tags following [semver pre-release naming](https://semver.org/#spec-item-9) will only update the full version (see https://github.com/docker/metadata-action?tab=readme-ov-file#typesemver).

Also updated actions to their latest version and refactored build workflows.


.github/workflows/build.yml
```diff
@@ -0,0 +1,157 @@
...
+    env:
+      FRONTEND_ENV: ${{ secrets.FRONTEND_ENV }}
+
+    strategy:
+      matrix:
+        node-version: [16.x]
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use Node.js ${{ matrix.node-version }}
```
shepilov, line 99: What is matrix.node-version?

ericlinagora: The matrix is an input of different combinations to try to run on. For ex. you could have os: mac, linux and node: 18, 20, and the build will try all 4 combinations. https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs

PR #444, opened 2024-03-12
#50 user picker UI improvements

### LINKED ISSUE #50: User picker does not work well
**Describe the bug**
- [ ] The user picker in access management doesn't work very well and doesn't match the current UI. We should rebuild it from scratch.
- [ ] Also the users are not loaded until we open the picker
- [x] Also we will need to remove the anonymous user from user search

**To Reproduce**
Open account B and lookup a file access management created by user A, you will not see user A:

Before searching:
Screenshot 2023-05-11 at 10 58 02

After searching (user repository updated): 
Screenshot 2023-05-11 at 10 58 07

ericlinagora:
About improvements called for in #50 
- Show when user results are truncated
- More better layout and colours even in dark-mode
- Keyboard navigation
- Clarify rights to be added


shepilov:
Small UI/UX improvements:

1. Make the "edit" icon the same color as label
image

2. The "Copy link" button should be the same in green state as in blue
image

3. The "remove password" modal window should look like this, without red color and with this text and buttons order
image

4. There shouldn't be a border between input and select, also the background color should be the same
image

5. When a user is the owner of the file, he should be displayed in the list like this
image

6. For the search results, the selected item should have a grey hoover and all the area should be clickable
image

7. Reduce the margin between input and search results
image

8. Add the "Done" button to the modal window, this button only closes the window
image

9. Remove the red color everywhere for "no access"
image

ericlinagora:
- 1 Make the "edit" icon the same color as label
  - done
- 2 The "Copy link" button should be the same in green state as in blue
  - done - just set a big width hopefully big enough for future translations
- 3 The "remove password" modal window should look like this, without red color and with this text and buttons order
  - the dialog in the screen shot is about deleting the link not the password, ignoring the text. Changed order, removed icon, made button blue. There's also two confirmation dialogs. Translation keys are welcome.
- 4 There shouldn't be a border between input and select, also the background color should be the same
  - don't think there is one, already is the same
- 5 When a user is the owner of the file, he should be displayed in the list like this
  - We need a way to change the rights of the file creator though, so where to put it ? There isn't really a "owner" concept. Drawing is unclear, if user is owner, creator, or both, to disable editing rights. No answer to questions.
- 6 For the search results, the selected item should have a grey hoover and all the area should be clickable
  - done
- 7 Reduce the margin between input and search results
  - done, but because we are incapable of getting dimensions, just a guess
- 8 Add the "Done" button to the modal window, this button only closes the window
  - done
- 9 Remove the red color everywhere for "no access"
  - done

tdrive/frontend/public/locales/ru.json
```diff
@@ -180,25 +180,25 @@
...
-    "components.public-link-security_expiration_title": "Истечение срока действия ссылки",
+    "components.public-link-security_password": "Пароль",
+    "components.public-link-security_password_copied": "Скопировать пароль в буфер",
+    "components.public-link-security_expired": "Срок действия",
+    "components.public-link-security_expiration_title": "Срок действия",
     "components.public-link-security_expiration_removal_title": "О удалении срока действия истекшей ссылки",
-    "components.public-link-security_expiration_removal_no_password": "Предупреждение: о удалении срока действия общедоступной ссылки, что делает её доступной для любого, кто знает её",
-    "components.public-link-security_expiration_removal_but_password": "Предупреждение: о удалении срока действия общедоступной ссылки, что делает её доступной для любого, кто знает пароль",
+    "components.public-link-security_expiration_removal_no_password": "Это делает документ доступным для любого, у кого есть эта ссылка",
+    "components.public-link-security_expiration_removal_but_password": "Это делает её доступной для любого, кто знает пароль",
```
ericlinagora, line 189: If google translate serves me well, this looks like more than fixing the language but a rephrasing of it, if you can do the en too so it suits better, I'll then do the fr ?

shepilov (approved):

Issue #438, opened 2024-03-12
[Feature] Make possible to run CLI application for Twake Drive


shepilov:
- [x]  Try to run the existing application that we have
- [x]  Make it possible to run `tdrive/backend/node/src/cli/cmds/search_cmds/index_all.ts` to rebuild search index
- [x]  Update documentation, on how to build and run cli scripts
- [ ]  (OPTIONAL) Modify this rebuild process to make it possible to delete entries from the search index that had been deleted from the DB

ericlinagora:
Didn't find a good way to do the optional task: it would require a sorted complete scan of both data sources to match up superfluous entries

PR #463, opened 2024-03-17
First fix of backend CLI tool #438

ericlinagora:


tdrive/backend/node/src/cli/cmds/search_cmds/index_all.ts
```diff
@@ -1,142 +1,210 @@
...
+  pageSizeAsStringForReasons: string = defaultPageSize,
+  filter: FindFilter = {},
+  delayParPageMS: number = 200,
+) {
+  let page: Pagination = { limitStr: pageSizeAsStringForReasons };
+  do {
+    const list = await repository.find(filter, { pagination: page }, undefined);
+    await forEachPage(list.getEntities());
+    page = list.nextPage as Pagination;
+    await waitTimeoutMS(delayParPageMS);
```
shepilov: why do we need this timeout?

ericlinagora: I don't really know, it was there before so I left it, but it makes sense to not be too brutal with db so it has a chance of running against a live instance

tdrive/backend/node/src/cli/lib/print_config_summary.ts
shepilov, line 1: Please, use a dash in file names as a separator

tdrive/backend/node/src/cli/cmds/search_cmds/index_all.ts
```diff
@@ -1,142 +1,210 @@
...
+      for (const item of list) {
+        if (seen.has(item)) continue;
+        seen.add(item);
+        yield item;
+      }
     }
 
-    await migrator.run({
-      repository,
+    runWithPlatform("Re-index", async ({ spinner, platform }) => {
```
shepilov: why do we need this function wrapper runWithPlatform? It will be more readable to have a method that returns a platform, and the pass it to the constructor and run all the commans

tdrive/backend/node/src/cli/cmds/search_cmds/index_all.ts
```diff
@@ -1,142 +1,210 @@
...
-
-    const repository = (argv.repository || "") as string;
-
-    if (!repository) {
-      throw "No repository was set.";
+    const repositoryArg = argv[repositoryArgumentName];
+    const repositories =
+      typeof repositoryArg === "string" ? [repositoryArg] : (repositoryArg as [string]);
+
+    function* eachOnlyOnce<T>(list: Iterable<T>): Iterable<T> {
```
shepilov: it can be replaced with just `new Set()`

tdrive/backend/node/src/cli/cmds/search_cmds/index_all.ts
ericlinagora: Because it will be a standard wrapper for all the commands that use the platform. It deals not only with startup; but also shutdown; exit code, errors and is a nice single place to have so we can have this in common for all subcommands. I'm not sure how to make it more readable though `runWithPlatform( (platform) =>/* stuff using platform */))` seems pretty readable to me. It will also be the place we can decide on how we deal with the service list for example.

tdrive/backend/node/src/cli/cmds/search_cmds/index_all.ts
ericlinagora: I didn't think of that. In mean time I had already moved it out and used lodash's _.uniq, is that ok or should I still use a Set ? For the moment ordering doesn't matter where I use this (for the repositories but also for user_ids to filter)

shepilov (approved):

MontaGhanmy (approved):

PR #495, opened 2024-04-26
🐛 Fix files downloaded as untitled

### LINKED ISSUE #469: Files downloaded as 'Untitled'
**Describe the bug**

A file is shared with me (pure ASCII name). I want to download it, it is stored as Untitled-20240320-0827.docx

**To Reproduce**

Don't know how to reproduce, this is the first time is see that _(our internal Tdrive)_

**Expected behavior**

Don't change the name

**Desktop (please complete the following information):**
 - OS: Debian
 - Browser Firefox

MontaGhanmy:


tdrive/backend/node/src/services/documents/web/controllers/documents.ts
```diff
@@ -384,7 +384,7 @@ export class DocumentsController {
         archive.pipe(response.raw);
       } else if (archiveOrFile.file) {
         const data = archiveOrFile.file;
-        const filename = data.name.replace(/[^a-zA-Z0-9 -_.]/g, "");
+        const filename = encodeURIComponent(data.name.replace(/[^\p{L}0-9 -_.]/gu, ""));
```
ericlinagora: The `-` in the regex is wrong (not from your change but might as well). Because the `-` is between ` ` and `_` it will create a class with all ascii between those two. The fix is to put it at the end so `[^\p{L}0-9 _.-]` .

ericlinagora (approved):

tdrive/backend/node/src/services/documents/web/controllers/documents.ts
MontaGhanmy: Fixed @ericlinagora thank you!

PR #510, opened 2024-04-30
✨ Dependencies upgrade

MontaGhanmy:


tdrive/backend/node/package.json
```diff
@@ -30,7 +30,7 @@
     "test:unit:watch": "npm run test:unit -- --watchAll --verbose false | pino-pretty",
     "test:merge:json": "npx istanbul report --dir coverage/merged --include 'coverage/**/coverage-final.json' json-summary",
     "test:merge:text": "npx istanbul report --dir coverage/merged --include 'coverage/**/coverage-final.json' text > coverage/merged/coverage-report.txt",
-    "test:all": "jest test --forceExit --coverage --detectOpenHandles --testTimeout=300000 --verbose false --runInBand",
+    "test:all": "jest test --forceExit --coverage --detectOpenHandles --testTimeout=3000000 --verbose false --runInBand",
```
ericlinagora: Isn't that moving the timeout from 5m to 50m ? Maybe is it worth it also to flag the test that takes too long and make an issue to split it into shorter tests ?

tdrive/backend/node/package.json
```diff
@@ -48,14 +48,14 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/Tdrive/tdrive-backend.git"
```
ericlinagora, line 51: 👍

tdrive/backend/node/src/core/platform/services/storage/index.ts
```diff
@@ -115,6 +115,10 @@ export default class StorageService extends TdriveService<StorageAPI> implements
 
       const chunks = options?.totalChunks || 1;
       let count = 1;
+
+      //check that the file a really exists
+      await self._read(options?.totalChunks ? `${path}/chunk${count}` : path);
```
ericlinagora, line 120: Isn't `count` always `1` ? Does this mean that if `options?.totalChunks === null/undef` the file has a single chunk at the place there would a be a folder with all the chunks inside ? Why treat chunks==1 as special ?

tdrive/backend/node/src/core/platform/services/webserver/index.ts
```diff
@@ -138,7 +141,10 @@ export default class WebServerService extends TdriveService<WebServerAPI> implem
         res.type("text/html").send(stream);
       });
 
-      await this.server.listen(this.configuration.get<number>("port", 3000), "0.0.0.0");
+      await this.server.listen({
+        port: this.configuration.get<number>("port", 3000),
+        host: "0.0.0.0",
```
ericlinagora, line 146: Unrelated to PR: but maybe it's worth making the bind host configurable ? In cases eg. where it's behind a reverse proxy we explicitely do not want to bind to 0.0.0.0.. I can make an issue

ericlinagora:
PS: didn't actually try to download all the deps and run the e2e, I trust the CI

tdrive/backend/node/src/core/platform/services/webserver/index.ts
shepilov: How it's connected with a reverse proxy? Ok, we are binding it to all interfaces, what is the problem?

ericlinagora: Usually in prod you want a local nginx or similar, to take care of socket handling, piping static files, that stuff. In that case you want to prevent the application from being accessible directly, so as an additional layer, usually you only bind to localhost so only nginx can talk to you if you exposed the port somehow. It's safer that's why it's the default

tdrive/backend/node/src/core/platform/services/storage/index.ts
shepilov: Nope, I think It's not always one, for the cases when you uploaded a file with API without explicitly identifying a number of chunks, it will be null and in the "path" you will have a file and not a folder

tdrive/backend/node/src/core/platform/services/storage/index.ts
```diff
@@ -136,6 +140,7 @@ export default class StorageService extends TdriveService<StorageAPI> implements
           if (decipher) {
             stream = stream.pipe(decipher);
           }
+          callback(null, stream);
```
shepilov, line 143: @MontaGhanmy this row is also from my PR?)

MontaGhanmy: Yes @shepilov

tdrive/backend/node/src/core/platform/services/storage/index.ts
ericlinagora: It's set to 1 like 1 line above ? Cool thanks. Why not have just made 1 huge chunk so there's no special case ?

shepilov (approved):

tdrive/backend/node/package.json
shepilov: @MontaGhanmy  could you please revert it back, it also looks like a misprint from my PR

tdrive/backend/node/src/core/platform/services/storage/index.ts
shepilov: I don't know

tdrive/backend/node/src/core/platform/services/webserver/index.ts
shepilov: It can be only the case when you run drive not inside docker, nomatter vannila or k8s. and I'm not sure that in 2024 anybody run anything not inside docker. Except for the local development, and if we will add this configuration we need not to forget about it. 
We can have this issue but will be deep deep inside backlog

tdrive/backend/node/package.json
MontaGhanmy: @shepilov It's not a misprint, it was deliberate because some tests were timing out. Should I still revert it back?

Issue #315, opened 2024-01-10
[Bug] When you follow a file link, the window for entering password is shown for a second 


tprudentova:
Env: tdrive.qa.lin-saas.com

1. Open Access management of a file 
2. Create a link not protected by a password
3. Follow that link 
4. A window for entering password is still shown for a second before the file is opened

https://github.com/linagora/twake-drive/assets/97684890/3cb43c01-da4d-4608-8991-10ef01f8c064

ericlinagora:
possibly related to #512

PR #511, opened 2024-04-30
💄 front: don't flash password prompt when opening a shared link without password protection

ericlinagora:
Fix for ticket #315


MontaGhanmy (changes requested):
With the current change the public link only worked for authenticated users. Else, the user is shown an error/gets redirected to login page.

MontaGhanmy (approved):

PR #513, opened 2024-05-06
♻️ Removed unused files and move unused encryption algorithm to co con…

shepilov:


ericlinagora (approved):
Note that we advertise the encryption algorithm (Taisiya asked me about for some documentation or design)

MontaGhanmy (approved):

Issue #433, opened 2024-03-11
[Bug] When we delete files that was uploaded by "anonymous" user, it doesn't appear in the trash


shepilov:
**To Reproduce**
1. User 1 sharer folder by the link
2. User2 uploads some files to this folder
3. User1 deletes these files
4. These files are nowhere

**Rules to implement**

- If `Alice` creates a shared folder (shared folder - to find `Alice` - is the deepest one shared), and:
  - `Anon` uploads file, `*` deletes it -> trash of `Alice`
  - `Bob` uploads file, `*` deletes it -> trash of `Bob`

PR #536, opened 2024-05-22
[WIP] 433 anonymously uploaded files should appear in trash of folder creator

ericlinagora:
#433 WIP

very very strict minimum mvp

### DIFF
--- a/tdrive/backend/node/src/services/documents/services/index.ts
+++ b/tdrive/backend/node/src/services/documents/services/index.ts
@@ -632,12 +632,12 @@ export class DocumentsService {
         this.logger.error({ error: `${error}` }, "Failed to grant access to the drive item");
         throw new CrudException("User does not have access to this item or its children", 401);
       }
-
+      const path = await getPath(item.parent_id, this.repository, true, context);
       const previousParentId = item.parent_id;
       if (
         (await isInTrash(item, this.repository, context)) ||
         item.parent_id === this.TRASH ||
-        (await getPath(item.parent_id, this.repository, true, context))[0].id === this.TRASH
+        path[0].id === this.TRASH
       ) {
         //This item is already in trash, we can delete it definitively
 
@@ -681,22 +681,19 @@ export class DocumentsService {
           const creator = await this.userRepository.findOne({ id: item.creator });
           if (creator.type === "anonymous") {
             const loadedCreators = new Map<string, User>();
-            const path = await getPath(
-              item.id,
-              this.repository,
-              true,
-              context,
-              async item => {
-                if (!item.creator) return true;
-                const user =
-                  loadedCreators.get(item.creator) ??
-                  (await this.userRepository.findOne({ id: item.creator }));
-                loadedCreators.set(item.creator, user);
-                return user.type !== "anonymous";
-              },
-              true,
-            );
-            const [firstOwnedItem] = path;
+            let firstOwnedItem: DriveFile | undefined;
+            for (let i = path.length - 1; i >= 0; i--) {
+              const item = path[i];
+              if (!item.creator) continue;
+              const user =
+                loadedCreators.get(item.creator) ??
+                (await this.userRepository.findOne({ id: item.creator }));
+              loadedCreators.set(item.creator, user);
+              if (user.type !== "anonymous") {
+                firstOwnedItem = item;
+                break;
+              }
+            }
             if (firstOwnedItem) {
               const firstKnownCreator = loadedCreators.get(firstOwnedItem.creator);
               const accessEntitiesWithoutUser = item.access_info.entities.filter(
@@ -704,6 +701,8 @@ export class DocumentsService {
               );
               item.access_info.entities = [
                 ...accessEntitiesWithoutUser,
+                // This is not functionally required, but creates an audit trace of what
+                // happened to this anonymously uploaded file
                 {
                   type: "user",
                   id: firstKnownCreator.id,
--- a/tdrive/backend/node/src/services/documents/services/index.ts
+++ b/tdrive/backend/node/src/services/documents/services/index.ts
@@ -15,6 +15,8 @@ import { hasCompanyAdminLevel } from "../../../utils/company";
 import gr from "../../global-resolver";
 import { DriveFile, TYPE } from "../entities/drive-file";
 import { FileVersion, TYPE as FileVersionType } from "../entities/file-version";
+import User, { TYPE as UserType } from "../../user/entities/user";
+
 import {
   DriveTdriveTab as DriveTdriveTabEntity,
   TYPE as DriveTdriveTabRepoType,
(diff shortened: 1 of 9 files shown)


tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -678,6 +684,53 @@ export class DocumentsService {
...
+              const accessEntitiesWithoutUser = item.access_info.entities.filter(
+                ({ id, type }) => type != "user" || id != firstKnownCreator.id,
+              );
+              item.access_info.entities = [
+                ...accessEntitiesWithoutUser,
+                {
+                  type: "user",
+                  id: firstKnownCreator.id,
+                  level: "manage",
+                  grantor: context.user.id,
```
ericlinagora, line 710: should be original item.creator ? (ie anonymous user), to create some trace ? But then we loose who did the deletion

ericlinagora: (rights shouldn't be needed for well rights since the creator will change)

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -678,6 +684,53 @@ export class DocumentsService {
       } else {
         //This item is not in trash, we move it to trash
         item.is_in_trash = true;
+        // Check item belongs to someone
+        if (item.creator !== context?.user?.id) {
+          const creator = await this.userRepository.findOne({ id: item.creator });
+          if (creator.type === "anonymous") {
+            const loadedCreators = new Map<string, User>();
+            const path = await getPath(
```
shepilov: You already have path, it was retrieved before you came to this if statement, in the previous one

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -678,6 +684,53 @@ export class DocumentsService {
...
+                  grantor: context.user.id,
+                },
+              ];
+              item.creator = firstKnownCreator.id;
+            } else {
+              // Move to company trash
+              item.parent_id = "trash";
+              item.scope = "shared";
+            }
+            this.repository.save(item);
```
shepilov: missing await

tdrive/backend/node/src/services/documents/services/index.ts
shepilov: yes, I don't think we need to modify access_entities, only for some strange cases when we remove access permission for the creator for this parent directory.
But let's forget about such kind of cases, as far as we are going to modify our permission model

tdrive/backend/node/src/services/documents/services/index.ts
ericlinagora: after oob discussion,  the getpath modifications to encourage early bailout are worth the tiny inefficiency

tdrive/backend/node/src/services/documents/services/index.ts
ericlinagora: after oob discussion,  keeping the permission addition for audit purposes. Not using the anonymous user id as there is only one per company anyway

tdrive/backend/node/test/e2e/common/user-api.ts
```diff
@@ -227,6 +231,21 @@ export default class UserApi {
     return directory;
   }
 
+  /** Run the provided callback using the specified bearer JWT token.
+   * //TODO: Warning: does not override calls using `this.api` have to discuss
+   */
+  async impersonateWithJWT<T>(jwt: string, cb: () => Promise<T>): Promise<T> {
```
shepilov, line 235: It's not needed to rollback to the old token.

For anonymous access you can crate a new instance of the UserApi in the test, and store the token permanently.
These callbacks just bring complexity

const anonymousUser = UserApi()
anonymousUser.loginWithToken()

ericlinagora: It's more complicated than that, it's what I wanted to discuss. The UserAPI class has a user object that wouldn't be valid with the other JWT, it also loads workspace and anonymous user stuff which we don't need, and it inits the Api class with the user, which then loads it's jwt on its own. I can't figure out what they're supposed to be but they don't match a user nor a session one to one. This is a hacky way around all that complexity.

shepilov: "This is a hacky way around all that complexity.", exactly, so please let's do not add additional. 
Actually it represents a user session in the browser, and sometimes it can be with "real" user or anonymous session.

So we can refactor the constructor of the class, and make some fields optional, but until this, let's KISS
anonymousUser.jwt = token.value;

ericlinagora: 🤮 that's worse ! at least my hack is clean and has no side effects (also I disagree it's complex, it's one callback that's [advised](https://en.wikipedia.org/wiki/Advice_(programming))), and the call is readable and descriptive. Modifying the field from outside on an object used in multiple tests....

Yes I want to refactor UserApi, and also, remove User-Api, but, we should talk first because I'm not sure about the actual responsibility scope we should aim for

tdrive/backend/node/test/e2e/documents/trash.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+    expect(createItemResult.scope).toEqual("personal");
+
+    expect(await getTrashContentIds("personal")).not.toContain(createItemResult.id);
+    const moveToTrashResponse = await currentUser.delete(createItemResult.id);
+    expect(moveToTrashResponse.statusCode).toEqual(200);
+    expect(await getTrashContentIds("personal")).toContain(createItemResult.id);
+    expect(await getTrashContentIds("shared")).not.toContain(createItemResult.id);
+  });
+
+  describe("deleting a file uploaded by an anonymous user should go to the sharers trash", () => {
```
shepilov: always? I thought it should depend on whare this file was uploaded.

ericlinagora: Not sure I'm understanding: it goes to the owner of the shared folder, so yes it depends on where it was uploaded ?

shepilov: "sharers trash" was not understandable for me without ', it thought it's a misprint from "shared"

ericlinagora: Right, yeah, maybe I should find a better way to write that... I want to avoid `'`'s just to make the CLI easier to use without escaping to filter the tests

tdrive/backend/node/test/e2e/documents/trash.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+      const anonymouslyUploadedDoc = await currentUser.impersonatePublicLinkAccessOf(publiclyWriteableFolder, () =>
+        currentUser.createDefaultDocument({
+          parent_id: publiclyWriteableFolder.id,
+          scope: "personal",
+        }));
+
+      const secondaryUser = await UserApi.getInstance(platform!);
+
+      const deletionToTrashResponse = await secondaryUser.delete(anonymouslyUploadedDoc.id);
+      expect(deletionToTrashResponse.statusCode).toBe(200);
```
ericlinagora, line 172: @shepilov This works, but can you confirm it's expected that secondaryUser has the permissions to do that ?

shepilov (approved):

PR #541, opened 2024-05-24
🐛 Fixed bug with downloading zip folder

shepilov:


tdrive/backend/node/test/e2e/common/user-api.ts
```diff
@@ -419,7 +419,7 @@ export default class UserApi {
   };
 
   async getFolderOKCheck(id: string | "root" | "trash" | "shared_with_me") {
```
ericlinagora: This looks unused, or if it is to be used, why not in this test ?

shepilov: yes, a good point, it should be

PR #554, opened 2024-05-29
🐛 Fixed bug with ability to log in with revoked session

shepilov:


tdrive/backend/node/src/services/console/clients/remote.ts
```diff
@@ -320,7 +323,8 @@ export class ConsoleRemoteClient implements ConsoleServiceClient {
     const sessionRepository = gr.services.console.getSessionRepo();
     const session = await sessionRepository.findOne({ sid: payload.claims.sid });
     if (session) {
-      await sessionRepository.remove(session);
+      session.revoked_at = new Date().getTime();
+      await sessionRepository.save(session);
```
ericlinagora, line 327: when are those rows cleaned up ?

MontaGhanmy (approved):

tdrive/backend/node/src/services/console/clients/remote.ts
shepilov: Right now nowhen, we'll think about it later, it can be cron job or event

PR #555, opened 2024-05-29
🐛 don't include deleted items in archives

ericlinagora:
Pending a 401 bug on local login and zip downloads:

- A shares a folder publicly
  - A adds file `abc`
- anonymous uses link to upload file `def`
- A deletes `def`
- A downloads folder as zip file
- Zip file includes `def` - but here's the kicker: it shouldn't because it was deleted in step 2 (or 3 if you count sub-steps too)


ericlinagora:
a moment of respect
image

shepilov (approved):

PR #568, opened 2024-06-07
🐛 Fixed bug with zipping all files in the folder

shepilov:


tdrive/backend/node/src/services/documents/utils.ts
```diff
@@ -351,25 +351,30 @@ export const addDriveItemToArchive = async (
...
+          child.id,
+          child,
+          archive,
+          repository,
+          context,
+          `${prefix || ""}${item.name}/`,
+        );
+      }
+      nextPage = items.nextPage?.page_token;
+    } while (nextPage > "");
```
ericlinagora: `string > ""` ? is this for some special coercion rules like `null` or `false` or `"0"` do something interesting with `> ""`

shepilov: `typeof nextPage!='undefined' && nextPage` doesn't look interesting, or for you do?

ericlinagora: why not just `while (nextPage)` ? or `!!nextPage` if you want to make the coercion explicit ?

ericlinagora (approved):
(assuming the `file.name` -> `item.name` is correct)

PR #570, opened 2024-06-20
onlyoffice rework

ericlinagora:
## Types of changes

- [ ] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.


tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/postgres/postgres.ts
```diff
@@ -270,21 +270,72 @@ export class PostgresConnector extends AbstractConnector<PostgresConnectionOptio
...
+      fieldName,
+      previousValue,
+      newValue,
+    );
+    const result = await this.executeRaw(updateQuery);
+    if (!result)
+      throw new Error(
+        `Error updating ${JSON.stringify(fieldName)} field of ${JSON.stringify(
+          entity["id"],
+        )} from ${JSON.stringify(previousValue)} to ${JSON.stringify(
```
shepilov, line 297: why do you need to stringify value? as well as field name

ericlinagora: I don't need to, but I like it, this happens when there's something unexpected, doing it this way reveals whitespace, weird chars, non strings, etc

tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/postgres/postgres.ts
```diff
@@ -270,21 +270,72 @@ export class PostgresConnector extends AbstractConnector<PostgresConnectionOptio
...
 
-  private async execute(query: Query) {
+  async atomicCompareAndSet<Entity, FieldValueType>(
+    entity: Entity,
+    fieldName: keyof Entity,
+    previousValue: FieldValueType,
+    newValue: FieldValueType,
+  ): Promise<{
+    didSet: boolean;
+    currentValue: FieldValueType | null;
```
shepilov, line 284: Do you need to return the current value every time? Why not just true/false

shepilov:
@ericlinagora there are changes in the connector that don't allow us to build it, are you sure that it will work and we can merge it)

tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/postgres/postgres.ts
ericlinagora: Because even though it's not the case today, for an atomic operation you might need the exact value at the replacement attempt moment, and this allows a single API (to be fixed if we need it, but the fix will be bellow the db service only)

ericlinagora:
> @ericlinagora there are changes in the connector that don't allow us to build it, are you sure that it will work and we can merge it)

This is the PR that I found out it was also linted, but I have uncommitted changes and would prefer not to commit that now. I must have forgotten a few files (remember, this commit was because I need to show *something*...). It build here, doesn't work because of token stuff, but runs fine

tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/postgres/postgres.ts
shepilov: difficult to read logs with all this JSON.stringify

tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/postgres/postgres.ts
shepilov: It won't give you a value of the same DB version as during the replacement attempt. As far as we don't have transactions, and we have mongoDB. Right now it is absolutely the same as a user make a select query just right after an unsuccessful CAS.
But even if you don't need it, you force user to have this additional select query.
Could you make it optional at least?

tdrive/connectors/onlyoffice-connector/src/lib/polled-thingie-value.ts
```diff
@@ -0,0 +1,98 @@
+import logger from './logger';
+
+const now = () => new Date().getTime();
+type NotUndefined<T> = T extends undefined ? never : T;
+
+/**
+ * Hold a value that is periodically obtained from a fallible process, and
+ * track update times and errors.
+ */
+export class PolledThingieValue<ValueType> {
```
shepilov, line 10: have you considered using rxjs?

tdrive/connectors/onlyoffice-connector/src/lib/polled-thingie-value.ts
```diff
@@ -0,0 +1,98 @@
...
+  private lastOkTimeMs: number | undefined;
+  private lastKoTimeMs: number | undefined;
+  protected lastValue: ValueType | undefined;
+  protected lastKoError: any;
```
shepilov, line 14: why do we need it, it's never read or used?

tdrive/connectors/onlyoffice-connector/src/services/api.service.ts
```diff
@@ -40,38 +46,38 @@ class ApiService implements IApiService {
...
-      return await this.axios.post(url, payload, {
+      return await axiosWithToken.post(url, payload, {
         headers: {
           ...headers,
         },
       });
     } catch (error) {
-      loggerService.error('Failed to post: ', error.message);
+      logger.error('Failed to post to Twake drive: ', error.stack);
       this.refreshToken();
```
shepilov, line 65: await is missing on purpose?

tdrive/connectors/onlyoffice-connector/src/services/api.service.ts
```diff
@@ -40,38 +46,38 @@ class ApiService implements IApiService {
...
     try {
-      return await this.axios.post(url, payload, {
+      return await axiosWithToken.post(url, payload, {
         headers: {
           ...headers,
         },
       });
     } catch (error) {
-      loggerService.error('Failed to post: ', error.message);
+      logger.error('Failed to post to Twake drive: ', error.stack);
```
shepilov, line 64: we do nothing if the token has expired and wasn't refreshed by poller, so if we missed with configuration we will still has the same problems

tdrive/connectors/onlyoffice-connector/src/lib/polled-thingie-value.ts
ericlinagora: I assume we're talking about the timestamps and last ko, the timestamps actually control wether the last result succeeded or not (this class considers `undefined` as potential success value too - in that case the helper functions need to be used with lastFailed/hasValue). The error is stored for eventually putting it in a health endpoint.

tdrive/connectors/onlyoffice-connector/src/services/api.service.ts
ericlinagora: I think so, that's not my code and it won't remain that way

tdrive/connectors/onlyoffice-connector/src/services/api.service.ts
ericlinagora: if we miss configure, then yes, it won't respond succesfuly to requests, and will complain in the logs. It's not possible to distinguish "misconfigured" from "host not available right now"

shepilov (approved):

PR #579, opened 2024-07-16
✨ Ignore errors during nextcloud migration to be able to import files partially

shepilov:


tdrive/backend/utils/nextcloud-migration/test/executor.test.ts
```diff
@@ -0,0 +1,44 @@
+import { describe, expect, test } from "@jest/globals";
+import { FunctionExecutor } from "../src/executor";
+
+//FOR LOCAL DEBUG PURPOSE ONLY, ITS NOT A TEST
```
ericlinagora: then maybe `describe.skip` ?

shepilov: why skip?

shepilov: oh, I understood why, not skip, delete comment

ericlinagora (dismissed):

Issue #562, opened 2024-06-05
With S3 storage, if `STORAGE_S3_HOME_DIRECTORY != "tdrive"` files aren't uploaded nor downloaded correctly


ericlinagora:
**Describe the bug**
Using the following environment (or any environment with S3 and without `STORAGE_S3_HOME_DIRECTORY=tdrive`):

```sh
ACCOUNTS_TYPE=internal
STORAGE_DRIVER=S3

STORAGE_S3_BUCKET=tdrive-dev
STORAGE_S3_ENDPOINT=localhost
STORAGE_S3_ACCESS_KEY=minio
STORAGE_S3_SECRET_KEY=miniopassword

# including for completeness only:
DB_DRIVER=mongodb
SEARCH_DRIVER=mongodb
PUBSUB_TYPE=local

```

- Uploads seem to work from the UI perspective, but the [during_upload.log](https://github.com/user-attachments/files/15589245/during_upload.log) show `S3Error`s with the following file path not found:
```
tdrive-dev/files/d56f4e70-2342-11ef-8d0f-238afbd3c611/d5611da0-2342-11ef-8d0f-238afbd3c611/06054f26-89ee-442b-9cf1-73d168d4519f/chunk1
```

There is however a file in the FS of the bucket at:
```
tdrive-dev/tdrive/files/d56f4e70-2342-11ef-8d0f-238afbd3c611/d5611da0-2342-11ef-8d0f-238afbd3c611/06054f26-89ee-442b-9cf1-73d168d4519f/chunk1
```

> Note the missing `/tdrive/` in the logs

- Attempts to download result in a 500 error `{"statusCode":500,"error":"Internal Server Error","message":"Failed to download file"}`. Logs: 
[during_download_attempt.log](https://github.com/user-attachments/files/15589528/during_download_attempt.log)

ericlinagora:
Issue goes away with `STORAGE_S3_HOME_DIRECTORY=tdrive`, but reappears with `STORAGE_S3_HOME_DIRECTORY=plop`: 
[during_upload_with_plop_home.log](https://github.com/user-attachments/files/15590934/during_upload_with_plop_home.log)

It searches for:
```
plop/files/d56f4e70-2342-11ef-8d0f-238afbd3c611/d5611da0-2342-11ef-8d0f-238afbd3c611/bc277481-05ad-4055-b8b7-b7846608ea67/chunk1
```
[...]

PR #583, opened 2024-07-18
🐛 When we are creating path for the files to store them in S3

shepilov:
sometimes it's with trailing "/" and sometimes without


ericlinagora:
Fixes #562

ericlinagora (approved):

PR #621, opened 2024-07-23
Give a more appropriate default file name for zip file downloads #600

### LINKED ISSUE #600: [BUG] Can't create zip file when downloading multiple files
**Describe the bug**

When trying to download multiple files, a binary is generated not a zip file.
Test on chrome https://drive.stg.lin-saas.com/

**To Reproduce**
- Select multiple files
- download them
- a binary file named "zip" is downloaded

**Expected behavior**

A zip file is downloaded (ex: random_name.zip)

**Screenshots**

[Capture vidéo du 22-07-2024 11:12:45.webm](https://github.com/user-attachments/assets/838fd800-83d6-42ba-a1f9-e224eae96d00)

ericlinagora:
Files are now `twake_drive.zip` (a more complicated scheme could be come up with, use ai, number of files, types, etc, but it's better than nothing for now).

Note: Includes changes from PR #619

### DIFF
--- a/tdrive/backend/node/src/services/documents/web/controllers/documents.ts
+++ b/tdrive/backend/node/src/services/documents/web/controllers/documents.ts
@@ -479,7 +479,7 @@ export class DocumentsController {
       ids = items.children.map(item => item.id);
     }
 
-    if (isDirectory) {
+    if (isDirectory === true) {
       const items = await globalResolver.services.documents.documents.get(ids[0], context, true);
       ids = items.children.map(item => item.id);
     }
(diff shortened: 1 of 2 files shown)


tdrive/backend/node/src/services/documents/web/controllers/documents.ts
```diff
@@ -479,13 +479,14 @@ export class DocumentsController {
       ids = items.children.map(item => item.id);
     }
 
-    if (isDirectory) {
```
rumata-feathers, line 482: not sure why you need that, but sure

rumata-feathers (approved):
good job!

tdrive/backend/node/src/services/documents/web/controllers/documents.ts
ericlinagora: yeah that's from the other PR I depend on (#619), we'll find out in staging I guess =)

Issue #603, opened 2024-07-22
[Bug] Open file in new window


Te-Z:
**Describe the bug**
Tested on Chrome https://drive.stg.lin-saas.com/

Open in new window just opens a new window of Drive but does not open file in a new window

[Capture vidéo du 22-07-2024 11:21:54.webm](https://github.com/user-attachments/assets/69108c35-0762-4aed-9522-717899dcb44e)

shepilov:
let's just disable this function until we will fix preview routing, and it's not quickly

Issue #618, opened 2024-07-22
[Bug] Infinite loop when opening folder in new window


Te-Z:
Tested on Chrome https://drive.stg.lin-saas.com/

[Capture vidéo du 22-07-2024 15:27:21.webm](https://github.com/user-attachments/assets/e78eb2a4-3082-4be5-abde-c15fa1304b0d)

PR #624, opened 2024-07-24
💄 frontend: hide open in new window context menu (#603)

ericlinagora:
This just hides the context menu option "Open in new window", as a temporary patch until we actually fix the underlying issue in #603


ericlinagora:
Also hides #618

shepilov (approved):

PR #635, opened 2024-07-25
🚑️ frontend: opera didnt permit folder uploads (#630)

### LINKED ISSUE #630: [Bug] Cannot upload folders in Opera
1.0.3, Opera; https://drive.stg.lin-saas.com/

1. Open Drive in Opera
2. Click Upload -> Folders
3. You can't upload a folder

https://github.com/user-attachments/assets/41b67cbb-339d-44b6-9fa2-2a363738bbad

ericlinagora:
fixes #630

### DIFF
--- a/tdrive/frontend/src/app/components/uploads/upload-zone.tsx
+++ b/tdrive/frontend/src/app/components/uploads/upload-zone.tsx
@@ -46,6 +46,12 @@ export default class UploadZone extends React.Component<PropsType, StateType> {
       this.file_input.multiple = this.props.multiple ? true : false;
       this.file_input.directory = this.props.directory ? true : false;
       this.file_input.webkitdirectory = this.props.directory ? true : false;
+      /* The following seems redundant with the previous, but is required for
+       * Opera support, as for historic reasons, the properties on the
+       * DOMElement don't exactly match up 1 to 1 with the attributes.
+       */
+      if (this.props.directory)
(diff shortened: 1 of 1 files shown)


tdrive/frontend/src/app/components/uploads/upload-zone.tsx
```diff
@@ -46,6 +46,8 @@ export default class UploadZone extends React.Component<PropsType, StateType> {
       this.file_input.multiple = this.props.multiple ? true : false;
       this.file_input.directory = this.props.directory ? true : false;
       this.file_input.webkitdirectory = this.props.directory ? true : false;
+      if (this.props.directory)
```
shepilov, line 53: what is the difference with the previous line? are we sitting true first and then empty string?

ericlinagora: Because `domelement.propertyname = 'abc'` doesn't exactly match to `domelement.setAttribute('propertyname', 'abc')`, and opera shows it more often because the whole property thing was an internet explorer invention that opera was the first to copy, kind of, without specs (much like `document.all`, opera was first to reproduce it, that's why opera worked better than mozilla etc with websites made for IE, which was, all of them). The difference between true and empty string is because some properties (like `disabled` or `checked` or `webkitDirectory` are true or not by just their presence, to set to false you would have to `.removeAttribute()`

shepilov (approved):

Issue #573, opened 2024-06-24
Mountable Drive, local storage and synchronisation


Facyla:
**Is your feature request related to a problem? Please describe.**
Internet access is not always given, and there are plenty situations where other means that online web browser access to access Drive files and edit them, including transportation (train, flight, boat, underground, inside some buildings), low connection zones (most places besides cities), intermittent connectivity...
And also for sobriety reasons related to bandwith use, and the convenient ability to use whatever authoring tool one prefers.
Plus the ability to store / backup files for offline access. 
And also the resilience to maintain other copies of a file than a single online version.

**Describe the solution you'd like**
Files offline access (cf. #222 ) and synchronization, just like services like Dropbox, OwnCLoud/NextCloud or other wellknown drives provide, with selective synchronization folders.

**Describe alternatives you've considered**
A degraded alternative would be to provide an interface that enables local access to the drive content, and mount the "Drive" as a local drive, so users can access the files in the local filesystem. 
Implementation of this could be WebDAV, Samba or other similar online-drive access.
But it's less interesting than the ability to enable offline access and sync files.

shepilov:
Hi @Facyla, sorry, I didn't get the point, what is the difference for you between "sync like Dropbox" and "mounting remote folder with WebDav" ?

Facyla:
sync like dropbox keeps the files locally when connectivity is not available (so we have 2 versions, the inline one and the local one, which are synced from time to time), 
while webDAV is an online shared drive, that is not available if connectivity is not available (so there is only a single version of the file, which may become unavailable).

However, the "dropbox-like" behaviour can be built using based on a shared drive mount (basically i would use a scripted rsync on a shared drive to mimic the dropbox behaviour).

PR #643, opened 2024-08-27
✨ Webdav connector to the Twake Drive

rumata-feathers:
WebDAV implementation

## Description
This is implementation the WebDAV protocol for our application using the nephele library.
Key components of this implementation include:
- Creating new 'webdav' service
- Custom adapter class to interface
- Resource class wrapper for drive items
- Device authorisation
- Implementing locks for the drive items

## Related Issue
https://github.com/linagora/twake-drive/issues/573

## Motivation and Context
WebDAV protocol is needed for desktop synchronisation

## How Has This Been Tested?
1. Manual testing using curl requests to simulate various WebDAV operations
2. Practical testing by mounting the WebDAV server as a network drive:
   - Tested on macOS Sonoma
   - Tested file creation, deletion, and modification through the mounted drive
  
Testing environment:
  - Native explorer
  - ACCOUNTS_TYPE=internal

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.


tdrive/backend/node/src/services/documents/entities/drive-file-lock.ts
```diff
@@ -0,0 +1,58 @@
...
+import { Column, Entity } from "../../../core/platform/services/database/services/orm/decorators";
+
+export const TYPE = "drive_file_locks";
+
+@Entity(TYPE, {
+  globalIndexes: [["company_id", "drive_file_id"], ["token"]],
+  primaryKey: [["company_id"], "id"],
+  type: TYPE,
+})
+export class DriveLock {
```
chibenwa, line 11: Sorry to be curious but could you explain me what a "lock" is in this context? What is it used for? 
Is this mandated by DAV?

tdrive/backend/node/src/services/documents/entities/drive-file-lock.ts
```diff
@@ -0,0 +1,58 @@
...
+  @Column("token", "string", { generator: "uuid" })
+  token: string;
+
+  @Type(() => Date)
+  @Column("created_at", "number")
+  created_at: number;
+
+  @Type(() => Number)
+  @Column("timeout", "number")
+  timeout: number;
```
chibenwa, line 34: A duration without unit? at least in a comment...

tdrive/backend/node/src/services/documents/entities/drive-file-lock.ts
```diff
@@ -0,0 +1,58 @@
...
+  @Column("drive_file_id", "uuid")
+  drive_file_id: string;
+
+  @Type(() => String)
+  @Column("token", "string", { generator: "uuid" })
+  token: string;
+
+  @Type(() => Date)
+  @Column("created_at", "number")
+  created_at: number;
```
chibenwa, line 29: I understand this is a date stored a s a unix timestamp. What is the timezone?
Also we do not have more expressive types for dates offered by the DB driver?

tdrive/backend/node/src/services/documents/entities/drive-file.ts
```diff
@@ -118,6 +119,9 @@ export class DriveFile {
   @Type(() => String)
   @Column("scope", "string")
   scope: DriveScope;
+
+  @Column("locks", "encoded_json")
+  locks: DriveLock[];
```
chibenwa, line 141: We add a column to an existing schema. We shall document the SQL command to add it in production data. Ops will <3 you for taking this care.

tdrive/backend/node/src/services/user/entities/device.ts
```diff
@@ -9,12 +10,18 @@ export const TYPE = "devices";
   type: TYPE,
 })
 export default class Device {
-  @Column("id", "string")
+  @Column("id", "uuid", { generator: "uuid" })
```
chibenwa: How do we migrate existing data?
Is changing type of a column allowed on a production system?

tdrive/backend/node/src/services/documents/entities/drive-file.ts
ericlinagora: drive has a *special* ORM, that just adds them itself if they're missing

(yes actual migrations are a current topic)

tdrive/backend/node/src/services/user/entities/device.ts
```diff
@@ -9,12 +10,18 @@ export const TYPE = "devices";
...
   id: string;
 
+  @Column("password", "string", { generator: "uuid" })
+  password: string;
```
chibenwa, line 28: Is this random enough? For such a sensible use case I would make sure I use cryptography grade random with enough entropy.

Also: does it make sense to store a password in clear? What's the impact if it gets read?

If it is stored only for validation later shouldn't we use bcrypt and friends?

tdrive/backend/node/src/services/webdav/web/adapter.ts
```diff
@@ -0,0 +1,223 @@
...
+export function getAdapterService() {
+  if (!AdapterService) {
+    throw new Error("AdapterService is not yet initialized");
+  }
+  return AdapterService;
+}
+export const adapterServiceReady = initializeAdapterService();
+
+export const getDriveExecutionContext = (url: URL): DriveExecutionContext => ({
+  //TODO[ASH] username will be actually a device, and we will need to get real user by device
```
chibenwa: Still relevant? Remove TODO?

tdrive/backend/node/src/services/user/entities/device.ts
ericlinagora: this, the *special* ORM doesn't do I think... but that table should be empty for the moment, this is the first actual use, we'll have to remember to test this before deploy

chibenwa (commented):
Partial review.
Disclaimer: I do not know tdrive, I do not know dav.
Feel free to discard comments if irrelevant.
Also: no tests? I see complex logic all across the place and many way things can go bust. It would be invaluable to invest in HTTP layer tests exercising this brand new DAV extension.
(from my experience standards needs **extensive** tests)

tdrive/backend/node/src/services/user/entities/device.ts
ericlinagora: I think we might need a clear text to support digest authentication (if realm is configurable).

The UUID should have enough entropy for crypto https://github.com/uuidjs/uuid

tdrive/backend/node/src/services/documents/entities/drive-file-lock.ts
ericlinagora: But unix timestamps don't need timezones, they are an instant in time past after 1/1/1970 UTC. We use a custom ORM that was supposed to be compatible with postgres, cassandra and mongo, and for that purpose is super generic. Something to fix, but out of scope of this PR

tdrive/backend/node/src/core/platform/services/webserver/index.ts
```diff
@@ -116,7 +116,7 @@ export default class WebServerService extends TdriveService<WebServerAPI> implem
       },
     });
     this.server.register(formbody);
-    this.server.register(corsPlugin, this.configuration.get<FastifyCorsOptions>("cors", {}));
+    // this.server.register(corsPlugin, this.configuration.get<FastifyCorsOptions>("cors", {}));
```
ericlinagora: Am I reading this right ? Does this disable CORS site-wide ????

tdrive/backend/node/src/services/documents/entities/drive-file.ts
ericlinagora, line 141: If DriveLocks are stored in that json field, why create the entity and db table and all that ?

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -1317,4 +1328,157 @@ export class DocumentsService {
       throw new CrudException(`Not enough space: ${size}, ${leftQuota}.`, 403);
     }
   };
+
+  /**
+   * Returns the file
```
ericlinagora, line 1767: I dont think this kind of documentation really adds anything ("returns the file" it's not hard to see the return type). However it would be nice, also eg on exported things like DriveLock

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -1317,4 +1328,157 @@ export class DocumentsService {
...
+      }
+    } catch (error) {
+      this.logger.error({ error: `${error}` }, "Failed to grant access to the drive item");
+      throw new CrudException("User does not have access to this item or its children", 401);
+    }
+    return Promise.resolve(driveFile);
+  };
+
+  /**
+   * Copies item to destination
```
ericlinagora, line 1816: ...

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -1317,4 +1328,157 @@ export class DocumentsService {
...
+      );
+      if (!hasAccess) {
+        this.logger.error("user does not have access drive item " + driveFile);
+        throw Error("user does not have access to this item");
+      }
+    } catch (error) {
+      this.logger.error({ error: `${error}` }, "Failed to grant access to the drive item");
+      throw new CrudException("User does not have access to this item or its children", 401);
+    }
+    return Promise.resolve(driveFile);
```
ericlinagora, line 1812: No need for `Promise.resolve` in an async function

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -1317,4 +1328,157 @@ export class DocumentsService {
...
+   */
+  copy = async (
+    id: string,
+    file: DriveFile | null,
+    content: Partial<DriveFile>,
+    context: DriveExecutionContext,
+  ): Promise<void> => {
+    assert(id != null, "Cannot copy null");
+    const item = await this.get(id, context);
+    assert(item.item.parent_id != null, "Cannot copy root");
```
ericlinagora, line 1830: what if `id` is not found ? How do you return say a 404 ? What about the asserts, shouldn't a lot of them be 400/401 type errors ?

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -1317,4 +1328,157 @@ export class DocumentsService {
...
+      "User does not have access to this item",
+    );
+    // TODO: properly check name
+    assert(content.parent_id != null && content.name != "root", "Cannot copy parent");
+    assert(
+      await checkAccess(content.parent_id, null, "write", this.repository, context),
+      "User does not have access to this item",
+    );
+    // now check that we can force copy if the item exists
+    // TODO: keep it or "write" level-access to parent allows to change any child?
```
ericlinagora, line 1842: @shepilov I think this is related to inheritance of rights ?

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -1317,4 +1328,157 @@ export class DocumentsService {
...
+  };
+
+  /**
+   * Moves item to destination
+   * @param id
+   * @param file
+   * @param content
+   * @param context
+   */
+  move = async (
```
ericlinagora, line 1879: Comments from `copy` apply

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -1317,4 +1328,157 @@ export class DocumentsService {
...
+    context: DriveExecutionContext,
+  ): Promise<void> => {
+    assert(id != null, "Cannot move null");
+    const item = await this.get(id, context);
+    assert(item.item.parent_id != null, "Cannot move root");
+    assert(
+      await checkAccess(id, file, "manage", this.repository, context),
+      "User does not have access to this item",
+    );
+    assert(content.parent_id != null && content.name != "root", "Cannot move to root");
```
ericlinagora, line 1892: why not ?

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -1317,4 +1328,157 @@ export class DocumentsService {
...
+   * Moves item to destination
+   * @param id
+   * @param file
+   * @param content
+   * @param context
+   */
+  move = async (
+    id: string,
+    file: DriveFile | null,
+    content: Partial<DriveFile>,
```
ericlinagora, line 1882: it's really not clear from the argument names what's what... The content has the new parent_id of the destination ? id is the source file ?

tdrive/backend/node/src/services/user/entities/device.ts
```diff
@@ -1,6 +1,7 @@
 import { merge } from "lodash";
 import { Column, Entity } from "../../../core/platform/services/database/services/orm/decorators";
 import { uuid } from "../../../utils/types";
+import {generators} from "openid-client";
```
ericlinagora: I think this is an IDE accident

tdrive/backend/node/src/services/user/services/users/service.ts
```diff
@@ -353,4 +378,24 @@ export class UserServiceImpl {
...
+    context?: ExecutionContext,
+  ): Promise<void> {
+    const user = await this.get(userPrimaryKey);
+    if (!user) {
+      throw CrudException.notFound(`User ${userPrimaryKey.id} not found`);
+    }
+    return await this.registerUserDevice(
+      userPrimaryKey,
+      randomUUID(),
+      "FCM",
```
ericlinagora: whats `FCM` ?

tdrive/backend/node/src/services/webdav/services/webdav.ts
```diff
@@ -0,0 +1,7 @@
+export class WebDAVServiceImpl {
```
ericlinagora, line 1: I dont get what this is doing ?

tdrive/backend/node/src/services/webdav/web/adapter.ts
```diff
@@ -0,0 +1,223 @@
...
+     *
+     * If the resource is not managed by this adapter, a BadGatewayError should be
+     * thrown.
+     *
+     * @param url Resource URL.
+     * @param baseUrl The root of the adapter's namespace on the server.
+     */
+    newResource = async (url: URL, baseUrl: URL): Promise<Resource> => {
+      const context = getDriveExecutionContext(url);
+      let pathname = decodeURI(url.pathname);
```
ericlinagora: doesn't really matter here, but in general you should always use `decodeURIComponent` not `decodeURI` (and the reciprocals). And you should really split by `/` first, then decode, in case of a filename containing `/`

tdrive/backend/node/src/services/webdav/web/adapter.ts
```diff
@@ -0,0 +1,223 @@
...
+     *
+     * @param url Resource URL.
+     * @param baseUrl The root of the adapter's namespace on the server.
+     */
+    newCollection = async (url: URL, baseUrl: URL): Promise<Resource> => {
+      const context = getDriveExecutionContext(url);
+      let pathname = url.pathname;
+      pathname = pathname.replace(baseUrl.pathname, "");
+      pathname = pathname.replace(/^\/+|\/+$/g, "");
+      const pathname_arr = pathname.split("/");
```
ericlinagora: at some point consider refactoring repeated code, not only does it document what's happening in the function name instead of having to read the code, but also if this logic ever needs to be tweaked/fixed, now we have to do it at many places, and sooner or later, a place will be forgotten

tdrive/backend/node/src/services/webdav/web/adapter.ts
```diff
@@ -0,0 +1,223 @@
...
+        throw new nephele.BadGatewayError("This resource is not managed by this adapter");
+      }
+      const resource = new ResourceService({
+        adapter: this,
+        baseUrl: baseUrl,
+        pathname: pathname_arr,
+        context: context,
+        is_collection: false,
+      });
+      return resource;
```
ericlinagora: Is it not an error if a resource at that path already exists ? More-so if permissions don't permit a write ?

tdrive/backend/node/src/services/webdav/web/adapter.ts
```diff
@@ -0,0 +1,223 @@
...
+     * returning specific error codes in certain situations, you should handle
+     * errors within this class' `run` function.
+     *
+     * If the requested method is not supported (i.e. it is purposefully excluded
+     * from the output from `getAllowedMethods`), a MethodNotSupportedError should
+     * be thrown. If the method is not recognized, a MethodNotImplementedError
+     * should be thrown.
+     */
+    getMethod = (method: string): typeof Method => {
+      if (method === "PROPATCH") {
```
ericlinagora: `PROP*P*ATCH`

but also, why, given the rest of the function ?

tdrive/backend/node/src/services/webdav/web/adapter.ts
```diff
@@ -0,0 +1,223 @@
...
+        throw new Error("Method not supported!");
+      }
+      throw new Error("Method not implemented.");
+    };
+  }
+  return new WebDAVAdapter();
+};
+
+const initializeAdapterService = async () => {
+  AdapterService = await eval("import('nephele').then(build_adapter)");
```
ericlinagora: I think it's worth a comment to document our shame, how `eval` keeps TS from transpiling `import` which wouldn't work with whatever module type combination it was

tdrive/backend/node/src/services/webdav/web/adapter.ts
```diff
@@ -0,0 +1,223 @@
...
+export const getDriveExecutionContext = (url: URL): DriveExecutionContext => ({
+  //TODO[ASH] username will be actually a device, and we will need to get real user by device
+  user: {
+    id: executionStorage.getStore().user_id,
+  } as types.User,
+  company: {
+    id: executionStorage.getStore().company_id,
+  },
+  url: url.href,
+  method: "undefined", // cannot retrieve method from url
```
ericlinagora: Yeah but that makes this a truthy string, I let it slide for the one above because it looked like it would be painful to change the type, but here ? ditto for req-id, make up a guid or whatever, ditto for transport, if we don't know, we don't (though I guess "insecure" is a good default)...

tdrive/backend/node/src/services/webdav/web/drivelock.ts
```diff
@@ -0,0 +1,198 @@
...
+      driveFile.locks = [];
+    }
+
+    // Check for conflicting locks
+    const conflictingLock = driveFile.locks.find(
+      lock =>
+        !lock.provisional &&
+        (this.scope === "exclusive" || lock.scope === "exclusive") &&
+        !DriveLock.isLockExpired(lock),
+    );
```
ericlinagora: Maybe I need to look into this more, but it seems provisional can't be ignored until expired, and also that you'd need to check parent items for locks with `.depth == "infinity"`, and if this lock is infinity, all child items for any locks ?

tdrive/backend/node/src/services/webdav/web/index.ts
```diff
@@ -0,0 +1,14 @@
...
+import { routes } from "./routes";
+
+export default async (
+  fastify: FastifyInstance,
+  options: FastifyRegisterOptions<{
+    prefix: string;
+  }>,
+): Promise<void> => {
+  fastify.log.debug("Configuring /webdav routes");
+  // const routes = eval ("import('nephele').then(builder)");
```
ericlinagora: needed comment ?

tdrive/backend/node/src/services/webdav/web/properties.ts
```diff
@@ -0,0 +1,290 @@
+import type { PropertyNotFoundError, Properties, User } from "nephele";
+import { ResourceService } from "./fileResource";
```
ericlinagora: our filename convention is `kebab-case`

tdrive/backend/node/src/services/webdav/web/properties.ts
```diff
@@ -0,0 +1,290 @@
...
+   *
+   * The following property is handled by Nephele.
+   *
+   * - lockdiscovery
+   *
+   * Any property not in the DAV: namespace will have its namespace and the
+   * string '%%' prepended to its name, like "LCGDM:%%mode".
+   */
+  get = async (name: string): Promise<string | object | object[] | undefined> => {
+    // const versions = await this.resource.getVersions();
```
ericlinagora: ?

tdrive/backend/node/src/services/webdav/web/properties.ts
```diff
@@ -0,0 +1,290 @@
...
+    switch (name) {
+      case "creationdate":
+        return Promise.resolve((this.resource.file.added || new Date().toISOString()).toString());
+      case "displayname":
+        return Promise.resolve(this.resource.getCanonicalName());
+      case "getcontentlanguage":
+        // TODO: keep it ?
+        return Promise.resolve("en");
```
ericlinagora: is it an issue if we don't have a value for properties like this ?

tdrive/backend/node/src/services/webdav/web/properties.ts
```diff
@@ -0,0 +1,290 @@
...
+   * - lockdiscovery
+   *
+   * Any property not in the DAV: namespace will have its namespace and the
+   * string '%%' prepended to its name, like "LCGDM:%%mode".
+   */
+  get = async (name: string): Promise<string | object | object[] | undefined> => {
+    // const versions = await this.resource.getVersions();
+    switch (name) {
+      case "creationdate":
+        return Promise.resolve((this.resource.file.added || new Date().toISOString()).toString());
```
ericlinagora: please remove Promise.resolve in async functions unless you need them for a good reason (not here)

tdrive/backend/node/src/services/webdav/web/properties.ts
```diff
@@ -0,0 +1,290 @@
...
+        return Promise.resolve(this.resource.getCanonicalName());
+      case "getcontentlanguage":
+        // TODO: keep it ?
+        return Promise.resolve("en");
+      case "getcontentlength":
+        return Promise.resolve(this.resource.file.size.toString());
+      case "getcontenttype":
+        const mediaType = await this.resource.getMediaType();
+        if (mediaType == null) {
+          return Promise.resolve("PropertyNotFoundError");
```
ericlinagora: nephele wants just that value returned and not an error thrown or something !?

tdrive/backend/node/src/services/webdav/web/properties.ts
```diff
@@ -0,0 +1,290 @@
...
+        }
+        return Promise.resolve(mediaType);
+      case "getlastmodified":
+        return Promise.resolve(
+          (this.resource.file.last_modified || new Date().toISOString()).toString(),
```
ericlinagora: @shepilov is it possible that `last_modified` is null but not `created_at` ?

If we don't have a value, we shouldn't just make up "now", this value's validity may be crucial for synchronisation software

tdrive/backend/node/src/services/webdav/web/properties.ts
```diff
@@ -0,0 +1,290 @@
...
+        };
+
+      case "getetag":
+        return this.resource.getEtag();
+      case "quota-available-bytes":
+        return `${await this.resource.getFreeSpace()}`;
+      case "quota-used-bytes":
+        return `${await this.resource.getTotalSpace()}`;
+      default:
+        throw new Error("PropertyNotFoundError") as PropertyNotFoundError;
```
ericlinagora: if nephele uses `instanceof`, which it probably does, this is going to break. The actual prototype has to be exported, from that place we do the `eval` I suppose

tdrive/backend/node/src/services/webdav/web/properties.ts
```diff
@@ -0,0 +1,290 @@
...
+   * - getlastmodified
+   * - resourcetype
+   * - supportedlock
+   *
+   * The following property is handled by Nephele.
+   *
+   * - lockdiscovery
+   */
+  set = (name: string, value: string | object | object[] | undefined): Promise<void> => {
+    return Promise.resolve();
```
ericlinagora: At least a `//TODO`, and maybe see how clients react to refusing this

ericlinagora: ditto for all the non implemented functions below

tdrive/backend/node/src/services/webdav/web/properties.ts
```diff
@@ -0,0 +1,290 @@
...
+   * property's entry in the object.
+   *
+   * The following property is handled by Nephele, and it is automatically
+   * included if your adapter supports locks, indicated by returning compliance
+   * class "2".
+   *
+   * - lockdiscovery
+   */
+  getAll = async (): Promise<{ [k: string]: string | object | object[] }> => {
+    const properties = [
```
ericlinagora: Note: We may need to store client-specific properties and retrieve them here, maybe worth a `//TODO` for the moment

tdrive/backend/node/src/services/webdav/web/properties.ts
```diff
@@ -0,0 +1,290 @@
...
+      "supportedlock",
+    ];
+    const result: { [k: string]: string | object | object[] } = {};
+    for (const property in properties) {
+      try {
+        result.property = await this.get(property);
+      } catch (e) {
+        result.property = e;
+      }
+    }
```
ericlinagora: Could be worth parallelising this, eg with `Promise.all(properties.map(....`

tdrive/backend/node/src/services/webdav/web/routes.ts
```diff
@@ -0,0 +1,92 @@
...
+    const authenticator = {
+      authenticate: async (request: express.Request, response: AuthResponse): Promise<User> => {
+        // console.log(request.headers, request.cookies, request.body, request.secret);
+        if (request.headers.authorization) {
+          // TODO: make auth just via login and password and not id's
+          try {
+            const base64Credentials = request.headers.authorization.split(" ")[1];
+            const credentials = Buffer.from(base64Credentials, "base64").toString("utf8");
+            const device_id = credentials.split(":")[0];
+            const device_password = credentials.split(":")[1];
```
ericlinagora: because there's no checks that neither `device_id` nor `device_password` have no `:` chars, it's better to split it once, then check you have exactly 2 results, at the very least we'll get a helpful error out of it

tdrive/backend/node/src/services/webdav/web/routes.ts
```diff
@@ -0,0 +1,92 @@
...
+            response.locals.user = {
+              username: device.user_id,
+              groupname: device.company_id,
+            } as User;
+            executionStorage.getStore().user_id = device.user_id;
+            executionStorage.getStore().company_id = device.company_id;
+            response.setHeader("WWW-Authenticate", "Basic");
+            return response.locals.user;
+          } catch (error) {
+            throw new nephele.UnauthorizedError("Error while authorising");
```
ericlinagora: not sure how nephele handles this, but this error could be anything. If you know it's the wrong user/pass, then fine, send a 401, but if you don't, send a 500, because when the client gets a 401, it will delete its stored credentials since it thinks them invalid, which mean a short network error, a restarting db, stuff like that and everyone has to re-enter their logins

tdrive/backend/node/src/services/webdav/web/routes.ts
```diff
@@ -0,0 +1,92 @@
...
+          throw new nephele.UnauthorizedError("Unauthorized user!");
+        }
+      },
+      cleanAuthentication: async (
+        request: express.Request,
+        response: AuthResponse,
+      ): Promise<void> => {
+        // TODO: think about cleaning the user
+        response.set("WWW-Authenticate", "Basic");
+        console.log("AUTHENTICATOR::cleanAuthentication is called()");
```
ericlinagora: no `console.log`s please

tdrive/backend/node/src/services/webdav/web/routes.ts
```diff
@@ -0,0 +1,92 @@
...
+      },
+    } as Authenticator;
+
+    await adapterServiceReady;
+    const adapter = getAdapterService();
+    fastify.register(fastifyExpress).after(() => {
+      // Create Nephele server
+      const server = nephele.createServer({
+        adapter: adapter, // You need to define this
+        authenticator: authenticator, // You need to define this
```
ericlinagora: I get this is example copy-pasta, but them comments....

tdrive/backend/node/src/services/webdav/web/routes.ts
```diff
@@ -0,0 +1,92 @@
...
+            res.status(500).send("Internal Server Error");
+          } else {
+            next();
+          }
+        });
+      });
+      fastify.use(`${webdavUrl}`, webdavMiddleware);
+    });
+    fastify.all("webdav/*", (request, reply) => {
+      console.log(request);
```
ericlinagora: .

tdrive/backend/node/src/services/webdav/web/routes.ts
```diff
@@ -0,0 +1,92 @@
...
+        plugins: {},
+      });
+
+      // Create an Express middleware that uses the Nephele server
+      const webdavMiddleware = express.Router();
+      webdavMiddleware.use(express.urlencoded({ extended: true }));
+      webdavMiddleware.use((req, res, next) => {
+        server(req, res, err => {
+          if (err) {
+            fastify.log.error("Nephele error:", err);
```
ericlinagora: please use our logger, `import { logger } from "./core/platform/framework/logger";` (path is from `src/`, you'll need to `../` it)

tdrive/backend/node/src/services/webdav/web/routes.ts
```diff
@@ -0,0 +1,92 @@
...
+            res.status(500).send("Internal Server Error");
+          } else {
+            next();
+          }
+        });
+      });
+      fastify.use(`${webdavUrl}`, webdavMiddleware);
```
ericlinagora: why not just `webdavUrl` ? Can it be a non-string ? (and if so, do we want this to work ?)

tdrive/backend/node/src/services/webdav/web/routes.ts
```diff
@@ -0,0 +1,92 @@
...
+    });
+    fastify.all("webdav/*", (request, reply) => {
+      console.log(request);
+      reply.send({ message: "Test route" });
+    });
+    fastify.all("webdav", (request, reply) => {
+      reply.send({ message: "Test route" });
```
ericlinagora: needed ?

tdrive/backend/node/src/services/webdav/web/routes.ts
```diff
@@ -0,0 +1,92 @@
...
+    });
+    console.debug(fastify.printRoutes());
```
ericlinagora: .

tdrive/backend/node/src/services/webdav/web/routes.ts
```diff
@@ -0,0 +1,92 @@
...
+    next();
+  };
+
+  return routes;
+}
+export const routes = eval("import('nephele').then(builder)");
```
ericlinagora: I'd really prefer if we did this horrible eval hack in a single place and exported what we need from there, and documented the raison-d'etre

tdrive/backend/node/src/utils/files.ts
```diff
@@ -81,6 +81,44 @@ export const readableToBuffer = async (input: Readable): Promise<Buffer> => {
...
+ *   the transformer stream and the resulting piped stream will be in
+ *   the result's `stream` property. If none is provided, that property
+ *   will be the transform stream the caller will need to pipe to themself.
+ * @returns An object with:
+ *    - `stream`: Either the result of `input.pipe(sizeAdderTransform)` if an input
+ *      was provided, or directly the stream the caller must pipe into.
+ *    - `getSize`: A function that can only be called when the stream is
+ *      finished, and returns the total size that was written into the
+ *      `sizeAdderTransform` stream.
+ */
```
ericlinagora, line 99: Ah, great inline documentation ! Personally I prioritise this over the separate documentation, makes it less likely to rot, and it's right there in the IDE

tdrive/backend/node/tsconfig.json
```diff
@@ -16,7 +16,7 @@
               "src/types/*"
           ],
           "src/*": ["src/*"]
-      }
+      },
```
ericlinagora: technically that's not valid json, seems benign, but also, why in this pr ?

ericlinagora (changes requested):

tdrive/backend/node/src/services/documents/services/index.ts
ericlinagora: Oh, because My Drive is not root..... I see.... this is awkward, maybe it should be and make company drive have another url for its root

tdrive/backend/node/src/services/webdav/web/adapter.ts
ericlinagora: does something like `global["im" + "port"]('nephele')` work ?

tdrive/backend/node/src/services/webdav/web/adapter.ts
ericlinagora: to please the gods of typing per microsoft, maybe `undefined as string` (or `undefined as unknown as string`), is the better option

tdrive/backend/node/src/core/platform/services/webserver/index.ts
shepilov: yes, it should be reconfigured only for the webdav uri

ericlinagora: i'm surprised webdav clients care at all about CORS

chibenwa: I had similar issues with JMAP

 - CORS do not makes sense for JMAP so *native* JMAP clients ignores them
 - But web based JMAP clients are bound to browser safety limitations hence CORS...

tdrive/backend/node/src/services/documents/services/index.ts
shepilov: yes, and it's more complicated than "keep it or "write" level-access to parent allows to change any child"

tdrive/backend/node/src/services/user/entities/device.ts
shepilov: @rumata-feathers, could you please take a look at how password is encoded in the "user" entity, and do the same with the same salt and the same passwordEncode component

tdrive/backend/node/src/core/platform/services/webserver/index.ts
ericlinagora: That makes sense, but still I'd prefer if native clients indeed ignore to keep them everywhere, and if/when we do have a browser client to support, and if it is a problem, then we look into removing them for that

tdrive/backend/node/src/services/user/services/users/service.ts
shepilov: football club monaco

tdrive/backend/node/src/services/webdav/services/webdav.ts
shepilov: probably it should do what adapter service is doing

ericlinagora: look at the whole file, if that's useful is purely by side-effect and worth a comment

tdrive/backend/node/src/services/webdav/web/properties.ts
shepilov: I can't imaging a situation when we have `created_at` but don't have `last_modified`, but if assume that everything is possible in Twake Drive, what is true meanwhile, yes, it should be `created_at` and only then `now()`

ericlinagora: I think it should never be `now()`, if we don't have a date, then tell the client we don't, inventing a date that is always changing would be far messier

tdrive/backend/node/src/services/user/services/users/service.ts
rumata-feathers: Firebase Cloud Messaging

tdrive/backend/node/src/services/webdav/web/adapter.ts
rumata-feathers: Not exactly, because newResource function in the adapter does not create the drive item, but the empty resource. All the checks are later done in the fileResource

tdrive/backend/node/src/services/user/services/users/service.ts
chibenwa: What is FCM used for in a web application?

ericlinagora: I got an offline answer: so this table represents devices that have ±delegate authentication for a user, and were used for notifications. I think it's legacy though in this case and we should explicitely set it to something indicating there are no notifications for webdav "device"s, @shepilov ?

shepilov: or just remove type field at all

tdrive/backend/node/src/services/webdav/web/routes.ts
```diff
@@ -76,14 +75,6 @@ async function builder(nephele: any) {
       });
       fastify.use(`${webdavUrl}`, webdavMiddleware);
     });
-    fastify.all("webdav/*", (request, reply) => {
```
ericlinagora: apparently this chunk is vital, without it, we get 404s only in the webdav stuff....

shepilov (approved):

shepilov (changes requested):
!

.github/workflows/build.yml
```diff
@@ -62,7 +62,7 @@ jobs:
       - name: e2e-mongo-s3-test
         run: |
           cd tdrive
-          docker compose -f docker-compose.tests.yml run --rm -e NODE_OPTIONS=--unhandled-rejections=warn -e SEARCH_DRIVER=mongodb -e DB_DRIVER=mongodb -e PUBSUB_TYPE=local node npm run test:all
+          docker compose -f docker-compose.tests.yml run --rm -e "NODE_OPTIONS=--unhandled-rejections=warn --experimental-vm-modules" -e SEARCH_DRIVER=mongodb -e DB_DRIVER=mongodb -e PUBSUB_TYPE=local node npm run test:all
```
ericlinagora: Still needed after the `eval` evil hack ?

tdrive/backend/node/src/core/platform/services/webserver/index.ts
shepilov: as far as I have seen in the WebDAV, it will be impossible to use web WebDAV client with CORS as far as they both use OPTIONS request to.
CORS support in fastify automatically answers this OPTIONS request with "Access-Control-Allow-Origin: " and hence doesn't pass further to WebDAV app.
So this support of the WebDAV for web clients will be limited to browser that a sending this "preflight" OPTIONS request and we will need to remove fastify on this path
So for now let's just  disable CORS for WebDAV URLs only

tdrive/backend/node/src/services/documents/entities/drive-file.ts
shepilov: Just naive copy-paste implementtion, anyway we will need to work more on lock

ericlinagora: yeah, but it does make sense to have them indexed in the db: when locking, we need to query locks above and below and on same item

ericlinagora:
rclone for testing webdav:
- copy url in twake drive account settings
- `rclone config` for interactive, pick webdav, vendor = other, no user/pass (it's in the url), lets say you called the remote `td`
- `rclone lsl td:` should list My Drive
- Debug flags: `-vv` for debug level, request debugging at https://rclone.org/docs/#dump-flag-flag-flag

The config file look like:

```
$ cat ~/.config/rclone/rclone.conf 
[td]
type = webdav
url = http://f2a30b0f-ac2f-4a87-b0d5-c718b0d3b0e2:53503dce-9e7b-4f2d-9bd0-7a894386ce27@localhost:4000/internal/services/webdav/v1/webdav/My%20Drive/
vendor = other
```

PR #646, opened 2024-09-14
🐛 backend: fix bug with postgres support when uploading files

ericlinagora:


tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -227,7 +227,7 @@ export class DocumentsService {
     if (options?.pagination) {
       const { page_token, limitStr } = options.pagination;
       const pageNumber =
-        dbType === "mongodb" ? parseInt(page_token) : parseInt(page_token) / parseInt(limitStr) + 1;
+        dbType === "mongodb" ? parseInt(page_token) : parseInt(page_token) / parseInt(limitStr);
```
ericlinagora, line 230: not a fan of having this dbtype specific outside of the adapters, and plus also that it doesn't throw a nice error if it's neither of two expected types

chibenwa (commented):
No associated tests?

ericlinagora:
> No associated tests?

We're looking into why this wasn't caught by current tests, because the scenario "i upload a file, i browse the folder and it's there" should really cover this.

MontaGhanmy (approved):

Issue #515, opened 2024-05-06
Remove trailing slash when reading configuration in the OnlyOffice connector


shepilov:
It's also a big mystery we should put the URL with or without a trailing slash in the URL confirmation for the connector.

So it's better to remove all slashes from the configuration and internally use URLs starting from slash

ericlinagora:
Most places I found remove any extraneous slashes before concatenating them, however the investigation revealed a series of other problems with the integration. I have not found a place where this was an issue, so marking as won't fix until more specific instructions for reproduction of the issue are added.

PR #650, opened 2024-09-18
onlyoffice rework

### LINKED ISSUE #523: Whe I edit document in OnlyOffice I don't see the changes made by another people and my changes are overrriden

### LINKED ISSUE #525: Too many version for the file are created during the file editing in the onlyOffice

### LINKED ISSUE #660: [BUG] Rename `folder or file` with space only in name
### Env: 

- Version: `1.0.4`
- Server: `stg.lin-saas.com`
- Platform: `Chrome & Microsoft Edge (Web, Mobile)`

### Steps to reproduce

1. Click “…” next to a file/folder
2. Select Rename
3. Enter space in name input field
4. Click `Update name` button.

## Expectation

Please trim it before checking empty

## Evidence

https://github.com/user-attachments/assets/e4e91c6e-6337-46c3-a5f5-4d49a38a35e9

ericlinagora:
#525 #515 #548 #523

This is a rewrite to have a more robust connexion with only office, recovering forgotten files, handling editing sessions as global locks, that kind


ericlinagora:
This is a bit of an express PR, so documentation will come later. As to testing most would require emulating the only office document server and the connector...

tdrive/backend/node/src/cli/cmds/editing_session_cmds/list.ts
```diff
@@ -0,0 +1,166 @@
...
+
+interface ListArguments {
+  all: boolean;
+  name: string;
+}
+
+async function report(platform: TdrivePlatform, args: ListArguments) {
+  const users = await makeUserCache(platform);
+  async function formatUser(id) {
+    const [wasKnown, user] = await users(id);
```
shepilov, line 39: how is it possible, that we have here "unknown" user_id?

ericlinagora: corrupt data, or maybe we invent virtual users kind of like the anonymous user today, either way, in a diagnostic tool, it's interesting if it happens

ericlinagora: ah, sorry I assumed it was about the `(user id not found)` below. The known or not here is wether it was in the cache, that way, the first time in the output we see that user, we get both email and id, then after we just get the email because the guid is annoying to read

tdrive/backend/node/src/cli/cmds/editing_session_cmds/list.ts
```diff
@@ -0,0 +1,166 @@
+import yargs from "yargs";
+
+import runWithPlatform from "../../lib/run-with-platform";
```
shepilov, line 3: could we also have some docs, and better also on the support page: how to use this command on production/stg envs

ericlinagora: This is really for debug, there's no limits or pagination etc, we should never run this outside of dev environments with nearly no data

tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/postgres/postgres-query-builder.ts
```diff
@@ -47,9 +47,18 @@ export class PostgresQueryBuilder {
               values.push(...inClause);
             }
           } else {
-            const value = `${this.dataTransformer.toDbString(filter, columnsDefinition[key].type)}`;
-            whereClause += `${key} = $${idx++} AND `;
-            values.push(value);
+            const isANotEqualFilter = filter && Object.keys(filter).join("!") === "$ne";
```
shepilov, line 50: Why did you implement it only for PostgreSQL, why do we need it in general, and why not use FindOptions.$nin instead?

ericlinagora: Because in mongo it's already the same syntax, so no need to. Postgres however wants the `is null` syntax, `not in` doesnt do what it looks like it should do (see https://stackoverflow.com/a/17877918 ).

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -958,41 +972,72 @@ export class DocumentsService {
...
+    const spinLoopUntilEditable = async (
+      provider: {
+        generateKey: () => string;
+        atomicSet: (
+          key: string | null,
+          previous: string | null,
+        ) => Promise<AtomicCompareAndSetResult<string>>;
+        getPluginKeyStatus: (key: string) => Promise<ApplicationEditingKeyStatus>;
+      },
+      attemptCount = 8,
```
shepilov, line 1011: magic number? and why do we need multiple attempts here in general? in what case? Only in case of unknown glitches?

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -958,41 +972,72 @@ export class DocumentsService {
...
+      tarpitS = 1,
```
shepilov, line 1012: tarpit?? reviewing you prs is always an English lesson for me

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -958,41 +972,72 @@ export class DocumentsService {
...
+      tarpitWorsenCoeff = 1.2,
+    ) => {
+      while (attemptCount-- > 0) {
+        const newKey = provider.generateKey();
+        const swapResult = await provider.atomicSet(newKey, null);
+        logger.debug(`Begin edit try ${newKey}, got: ${JSON.stringify(swapResult)}`);
+        if (swapResult.didSet) return newKey;
+        if (!swapResult.currentValue) continue; // glitch in the matrix but ok because atomicCompareAndSet is not actually completely atomic
```
shepilov, line 1020: Why it's not completely atomic?

shepilov, line 1020: If this glitch could happen let's log and monitor it, and we have event retries and tarpit for this

tdrive/backend/node/src/services/documents/services/index.ts
ericlinagora: https://en.wikipedia.org/wiki/Tarpit_(networking) =D

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -958,41 +972,72 @@ export class DocumentsService {
...
+        const existingStatus = await provider.getPluginKeyStatus(swapResult.currentValue);
+        logger.debug(`Begin edit get status of ${newKey}: ${JSON.stringify(existingStatus)}`);
+        switch (existingStatus) {
+          case ApplicationEditingKeyStatus.unknown:
+          case ApplicationEditingKeyStatus.live:
+            return swapResult.currentValue;
+          case ApplicationEditingKeyStatus.updated:
+          case ApplicationEditingKeyStatus.expired:
+            logger.debug(`Begin edit emptying previous ${swapResult.currentValue}`);
+            await provider.atomicSet(null, swapResult.currentValue);
```
shepilov, line 1030: and why not setting a new key here?

tdrive/backend/node/src/services/documents/services/index.ts
ericlinagora: In case of multiple people trying to start editing at the same time, first attempt tries to own it, then following attempts continue until settled. It's not that unlikely a scenario, "Hey team, I'll paste a link to a document we need to finish tonight because management fucked up the timing again"

tdrive/backend/node/src/services/documents/services/index.ts
ericlinagora: https://github.com/linagora/twake-drive/blob/main/tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/index.ts#L58

tdrive/backend/node/src/services/documents/services/index.ts
ericlinagora: (and magic - kind of, it does have a name/variable... I don't think this is worth putting in config, it's not like it needs to be different between one deployment and another, and we can still tweak in a centralised place)

tdrive/backend/node/src/services/documents/services/index.ts
ericlinagora: the `continue` is the retry, I don't see a point in logging it, it's 100% recoverable

tdrive/backend/node/src/services/documents/entities/drive-file.ts
```diff
@@ -120,6 +124,105 @@ export class DriveFile {
...
+    instanceId: string,
+    companyId: string,
+    userId: string,
+    overrideTimeStamp?: Date,
+  ) {
+    checkFieldValue("applicationId", applicationId);
+    checkFieldValue("instanceId", instanceId, false);
+    const isoUTCDateNoSpecialCharsNoMS = (overrideTimeStamp ?? new Date())
+      .toISOString()
+      .replace(/\..+$/, "")
```
shepilov, line 173: why so complicated? we are formating date, and removing parts after "." and timezone, could you please replace it just with your wished date format

tdrive/backend/node/src/services/documents/services/index.ts
ericlinagora: as the enum documents, in those cases, the plugin detected it was a forgotten file or other stale key, in this case the plugin already updated if possible the forgotten file in the getPluginKeyStatus call, or didn't for other reasons and it was logged as unknown, but the key is no longer valid, and should first be cleared in the drivefile, then a new key should be attempted by the first one to get there

tdrive/backend/node/src/services/documents/entities/drive-file.ts
```diff
@@ -120,6 +124,105 @@ export class DriveFile {
...
+      .replace(/[ZT:-]/g, "");
+    const userIdBuffer = UUIDTools.bufferFromUUIDString(userId) as unknown as Uint8Array;
+    const companyIdBuffer = UUIDTools.bufferFromUUIDString(companyId) as unknown as Uint8Array;
+    const entropyBuffer = UUIDTools.bufferFromUUIDString(randomUUID()) as unknown as Uint8Array;
+    const idsString = OnlyOfficeSafeDocKeyBase64.fromBuffer(
```
shepilov, line 178: and also why do we need all this tricky conversions, uuid to hex, and then back to uuid?

ericlinagora: Just using strings with available chars would be bigger than OO allows for key formats. Also because I want editing_session_keys to be sortable by timestamp lexicographically so we can batch work on the older ones

tdrive/backend/node/src/services/documents/entities/drive-file.ts
```diff
@@ -120,6 +124,105 @@ export class DriveFile {
...
+      )}). Must be short and only alpha numeric or '_' and '-'`,
+    );
+}
+/**
+ * Reference implementation for generating then parsing the {@link DriveFile.editing_session_key} field.
+ *
+ * Fields should be explicit, `instanceId` is for the case when we have multiple
+ * clients
+ */
+export const EditingSessionKeyFormat = {
```
shepilov, line 153: if you think it's still needed to have all this id manipulation, could you please provide a specific unit test for this class to make sure all conversions work as expected. 
And also it will be more understandable how it works when you see examples in test

ericlinagora: https://github.com/linagora/twake-drive/pull/650/files/630d2667d30cf86f9496f47722c474a465be40a4#diff-bb745da353734234ef85b7cadc5f3bc5cc6e15063f10ff02d4967a701cdfd8eb

tdrive/backend/node/src/services/documents/entities/drive-file.ts
ericlinagora: It's a compromise between between sorting chronologically being the same lexicographically, but also not caring about ms resolution and tz being always utc (Z), its to take less space in the key, it must be < 128 chars, there's a ts, an app, an instance, 2 guids we need, and 1 guid of entropy (we can maybe shorten but why ? it works)

tdrive/backend/node/src/services/documents/entities/drive-file.ts
ericlinagora: wrong link sorry https://github.com/linagora/twake-drive/pull/650/files/630d2667d30cf86f9496f47722c474a465be40a4#diff-16119ebf8045f3ebb7a272dee043d022a69a96623a1106c63bdfd32d9dcd9422

tdrive/backend/node/src/services/applications-api/index.ts
```diff
@@ -68,10 +86,95 @@ export default class ApplicationsApiService extends TdriveService<undefined> {
         }
       }
     }
-
+    ApplicationsApiService.default = this;
     return this;
   }
 
+  /** Get the configuration of a given `appId` or `undefined` if unknown */
+  public getApplicationConfig(appId: string) {
```
shepilov, line 94: It's better to user gr.services.applications.marketplaceApps.get() then config, later this application should be in the DB different for each company

tdrive/backend/node/src/services/applications-api/index.ts
```diff
@@ -68,10 +86,95 @@ export default class ApplicationsApiService extends TdriveService<undefined> {
...
+  /**
+   * Check status of `editing_session_key` in the corresponding application.
+   * @param editingSessionKey {@see DriveFile.editing_session_key} to check
+   * @returns status of the provided key as far as the application knows
+   */
+  async checkPendingEditingStatus(editingSessionKey: string): Promise<ApplicationEditingKeyStatus> {
+    const parsedKey = EditingSessionKeyFormat.parse(editingSessionKey);
+    const response = await this.requestFromApplication(
+      "POST",
+      "tdriveApi/1/session/" + encodeURIComponent(editingSessionKey) + "/check",
```
shepilov, line 150: hardcoded url? could you please put TODO with move it to application configuration later

tdrive/backend/node/src/services/applications-api/index.ts
ericlinagora: yes but today marketplaceApps has no apps

tdrive/backend/node/src/services/applications-api/index.ts
shepilov: or it should be implemented in every connector?

ericlinagora: what for ? if an project wants to be our plugin we can demand it responds to well known urls, and in this case this is the only place it is at. its never going to change

ericlinagora: yes every connector that uses an editing session key

ericlinagora: (if you're not the appId of an editing_session_key, this isn't getting called)

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -618,7 +621,17 @@ export class DocumentsService {
 
         await updateItemSize(oldParent, this.repository, context);
       }
-
+      if (renamedTo && item.editing_session_key)
+        ApplicationsApiService.getDefault()
+          .renameEditingKeyFilename(item.editing_session_key, renamedTo)
```
shepilov, line 626: If we can't rename a file in the application, should it prevent us of renaming the file in general? Should we just catch, log, and monitor?

tdrive/backend/node/src/services/documents/services/index.ts
shepilov: oh yeah, it's a current value that is not atomic at all

shepilov (approved):

PR #675, opened 2024-10-02
🐛 Fix URL encoded filename on download

### LINKED ISSUE #665: [Bug] If a file is named in Russian, it's downloaded with a weird name
1.0.4, drive.stg.lin-saas.com, tested on Safari, Opera, MacOS

1. Upload a file with a Russian name 
2. Download it
3. It has a really weird name

https://github.com/user-attachments/assets/90a2b32b-def3-46d1-bf9e-098e7675cfd2

MontaGhanmy:
The issue originated on **MacOS** platforms when testing on Safari and Opera browsers.

## Bug description
When users uploaded files with non-ASCII characters in their names (e.g., files with Russian characters), the filenames would become distorted upon download. This is a browser-related limitation where non-ASCII characters are not properly handled in the Content-Disposition header during file download.

Platform: MacOS, Browsers: Safari, Opera

## Fix description
We resolved this by updating the Content-Disposition header. The key here is the filename*= attribute. This is based on RFC 5987, which allows us to specify an encoding (in this case, UTF-8) to handle filenames that include special characters. Without filename*=, many browsers fall back to ASCII, causing filenames with non-ASCII characters to display incorrectly. By using this syntax, we ensure consistent handling of filenames across browsers and platforms.

This fix ensures proper cross-browser support, particularly when dealing with international file names or special characters, something that is crucial for user experience in globalized environments.

### DIFF
--- a/tdrive/backend/node/src/services/documents/web/controllers/documents.ts
+++ b/tdrive/backend/node/src/services/documents/web/controllers/documents.ts
@@ -24,6 +24,7 @@ import {
 import { DriveFileDTO } from "../dto/drive-file-dto";
 import { DriveFileDTOBuilder } from "../../services/drive-file-dto-builder";
 import config from "config";
+import { formatAttachmentContentDispositionHeader } from "../../../files/utils";
 
 export class DocumentsController {
   private driveFileDTOBuilder = new DriveFileDTOBuilder();
@@ -427,12 +428,8 @@ export class DocumentsController {
         return response;
       } else if (archiveOrFile.file) {
         const data = archiveOrFile.file;
-        const filename = encodeURIComponent(data.name.replace(/[^\p{L}0-9 _.-]/gu, ""));
 
-        response.header(
-          "Content-Disposition",
-          `attachment; filename="${filename}"; filename*=UTF-8''${filename}`,
-        );
+        response.header("Content-Disposition", formatAttachmentContentDispositionHeader(data.name));
 
         if (data.size) response.header("Content-Length", data.size);
         response.type(data.mime);
@@ -486,7 +483,10 @@ export class DocumentsController {
 
     try {
       const archive = await globalResolver.services.documents.documents.createZip(ids, context);
-      reply.raw.setHeader("content-disposition", 'attachment; filename="twake_drive.zip"');
+      reply.raw.setHeader(
+        "content-disposition",
+        formatAttachmentContentDispositionHeader("twake_drive.zip"),
+      );
 
       archive.on("finish", () => {
         reply.status(200);
--- a/tdrive/backend/node/src/services/documents/web/controllers/documents.ts
+++ b/tdrive/backend/node/src/services/documents/web/controllers/documents.ts
@@ -429,7 +429,11 @@ export class DocumentsController {
         const data = archiveOrFile.file;
         const filename = encodeURIComponent(data.name.replace(/[^\p{L}0-9 _.-]/gu, ""));
 
-        response.header("Content-disposition", `attachment; filename="${filename}"`);
+        response.header(
+          "Content-Disposition",
+          `attachment; filename="${filename}"; filename*=UTF-8''${filename}`,
+        );
+
         if (data.size) response.header("Content-Length", data.size);
         response.type(data.mime);
         return response.send(data.file);
--- a/tdrive/backend/node/src/services/files/web/controllers/files.ts
+++ b/tdrive/backend/node/src/services/files/web/controllers/files.ts
@@ -5,6 +5,7 @@ import { CompanyExecutionContext } from "../types";
 import { UploadOptions } from "../../types";
 import { PublicFile } from "../../entities/file";
 import gr from "../../../global-resolver";
+import { formatAttachmentContentDispositionHeader } from "../../utils";
 
 export class FileController {
   async save(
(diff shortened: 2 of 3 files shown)


tdrive/backend/node/src/services/files/web/controllers/files.ts
```diff
@@ -49,7 +49,11 @@ export class FileController {
       const data = await gr.services.files.download(params.id, context);
       const filename = data.name.replace(/[^a-zA-Z0-9 -_.]/g, "");
```
ericlinagora: - Why no `encodeURIComponent` on this one ? (why not centralise that code into a function)
- Why remove non latin chars, and whatever in ascii is between space and underscore, from filenames ? This works with the cyrillic filenames from the original issue ?

chibenwa: 100% agree, removing non ascii chars is **NOT** encoding.

chibenwa (commented):
Sorry, to me a **LOT** of things are going wrong here.

 - Remark 1: where are the tests?
 - Remark 2: UTF-8 / content type setting shall be handled by a library. It is likely to be way more robust.
  - Remark 3: pretty sure that with the current code we can inject content-type attributes EG if I have a file named `" charset="utf8`. Always beware of string concatenation and use specialized builders just like to prevent SQL injections.
  - Remark 4: For test/* document this piece of code looses the charset.

chibenwa (changes requested):
I do think more time needs to be spent on these subtle topics.

ericlinagora:
>     * Remark 1: where are the tests?

Testing this would require reimplementing the browser interpretation of this header for each browser*os, this bug was specific to one platform.

>     * Remark 2: UTF-8 / content type setting shall be handled by a library. It is likely to be way more robust.

This is content-disposition, to my knowledge our libraries don't support this correctly, It's better to just do the uri encode and RFC 5987 in a single place and re-use this

>     * Remark 3: pretty sure that with the current code we can inject content-type attributes EG if I have a file named `" charset="utf8`. Always beware of string concatenation and use specialized builders just like to prevent SQL injections.

Once we add the `encoreUriComponent` this should not be possible. And assuming fastify correctly handles multi-line header values, it shouldn't be possible to affect another header than content-disposition, which should really only affect the client's download anyway.

> 
>     * Remark 4: For test/* document this piece of code looses the charset.

I don't understand ?

chibenwa:
Remark 4: For text/* document this piece of code looses the charset.

s/test/text/g

chibenwa:
> Once we add the encoreUriComponent this should not be possible. And assuming fastify correctly handles multi-line header values, it shouldn't be possible to affect another header than content-disposition, which should really only affect the client's download anyway.

Do we have evidence of this?

Like a test?

chibenwa:
> Testing this would require reimplementing the browser interpretation of this header for each browser*os, this bug was specific to one platform.

I do not mean implement a complete test coverage for each browser but rather add one or two test to quickly see how the app behaves and cover edge cases if any.

If not just to document how filenames are transformed! We discovered the proposed change does some non obvious / controversial changes on the filename, changes which deserve to be explicitly stated in a test.

ericlinagora:
> Remark 4: For text/* document this piece of code looses the charset.
> 
> s/test/text/g

Ok, but again, this is not content-type, it's content-disposition. (besides I don't think we ever get the actual charset during the upload, we just restore as was)

> Do we have evidence of this?

There's a million assumptions we all make on our libs, we don't test everything. If fastify/express/node don't handle this correctly, it's a bug on their end. In terms of header injection there's nothing we could do beyond using the interface correctly and we do.

ericlinagora:
> > Testing this would require reimplementing the browser interpretation of this header for each browser*os, this bug was specific to one platform.
> 
> I do not mean implement a complete test coverage for each browser but rather add one or two test to quickly see how the app behaves and cover edge cases if any.
> 
> If not just to document how filenames are transformed! We discovered the proposed change does some non obvious / controversial changes on the filename, changes which deserve to be explicitly stated in a test.

All a test like that would do, would be to check we succesfully copied the expected from the actual into the test and look at the header. This issue fixes how safari interpreted this header, that much we can't test without emulating a browser. The removing chars in surely an error from legacy and we shouldn't document it, we should fix it.

chibenwa:
Well if this PR is open it is that there is a bit more than this ` fastify/express/node` mix in here but also logic written by the team, and if we get here it means that there was an issue.

Do you know about Test Driven Development?

A wonderful way to avoid regressions.

ericlinagora:
> Well if this PR is open it is that there is a bit more than this ` fastify/express/node` mix in here but also logic written by the team, and if we get here it means that there was an issue.
> 
> Do you know about Test Driven Development?
> 
> A wonderful way to avoid regressions.

dogmatism is a great way to increase tech debt and slow development vastly. Testing that url encode indeed url encodes and concatenating a string indeed concatenates is not a useful test. But be my guest and show me wrong with a useful test for a string replace.

tdrive/backend/node/src/services/documents/web/controllers/documents.ts
```diff
@@ -429,7 +429,11 @@ export class DocumentsController {
         const data = archiveOrFile.file;
         const filename = encodeURIComponent(data.name.replace(/[^\p{L}0-9 _.-]/gu, ""));
 
-        response.header("Content-disposition", `attachment; filename="${filename}"`);
+        response.header(
```
shepilov: @MontaGhanmy could you please describe by editing the very first comment to this issue what the bug was, for what platform, and why we need this filename*=?

shepilov (dismissed):

tdrive/backend/node/src/services/documents/web/controllers/documents.ts
MontaGhanmy: @shepilov I just did. It's updated.

shepilov (approved):

PR #682, opened 2024-10-08
💚 add healthchecks to the docker compose services used in CI testing  removed scylladb, add env vars that are always set, removed unused dependencies

ericlinagora:


ericlinagora:
In a previous version, I edited some backend files to trigger CI, it built and tested fine ( see https://github.com/linagora/twake-drive/actions/runs/11240298829/job/31249332049?pr=682 ) . I've removed those edits as irrelevant. No way to know this fixes the build issues, but it should be more correct regardless.

shepilov (approved):

PR #685, opened 2024-10-10
✨ Resilience strategy to read and write only to one storage

shepilov:


tdrive/backend/node/src/core/platform/services/storage/oneof-storage-strategy.ts
```diff
@@ -0,0 +1,115 @@
...
+   * @returns A readable stream of the file.
+   * @throws Error if all storage read attempts fail.
+   */
+  read = async (path: string, options?: ReadOptions): Promise<Readable> => {
+    for (const storage of this.storages) {
+      try {
+        if (await storage.exists(path, options)) {
+          return await storage.read(path, options);
+        }
+      } catch (err) {
```
ericlinagora, line 107: shouldn't this take care for ex. to respond 404 if all upstreams responded 404 (but they worked) for ex ? no rights issues possible etc ?

ericlinagora:
from the specs isn't this supposed to make some magic very recognisable log outputs when the backends don't agree ?

tdrive/backend/node/src/core/platform/services/storage/oneof-storage-strategy.ts
shepilov: maybe makes sense, right now we don't have this information anywhere even for one storage

shepilov: so an exception is thrown in StorageService on top of the read method and later should be watched in download FileService or controller methods

ericlinagora: with a super magic log because if this happens we have a big problem

PR #688, opened 2024-10-11
💄 Update the navigation bar/app grid

### LINKED ISSUE #689: [Improvement] [Feature] New twp navigation bar
The Twake Drive navigation bar and app grid have to be updated as described here: https://www.notion.so/linagora/TWP-navigation-bars-with-apps-grid-4861683b20e142ebaae6e7a348547652

MontaGhanmy:
## The updated navigation bar
image


shepilov:
@MontaGhanmy there shouldn't be firts/last name in the app bar

ericlinagora:
it's itchy to my OCD that the middle column isn't aligned with the app grid button, is this on purpose ?

shepilov:
yes, it should be aligned right border of app menu icon

ericlinagora:
Is this the alignment we wanted ? (not a fan, but if that's the design...)

image



I don't think is correct, the idea was to put the email in this dropdown as a disabled entry:

image

not here:

image

And bonus if you change that dialog you'll have a fun merge time come ~~webdav-express~~ the already merged addition of Vietnamese and dark mode for that dialog.

ericlinagora:
image

it's there and not clickable (my account in that screen shot has no first/last name but email is `a@b.c` ), should we show the e-mail as well ? Is there a way to make it obvious it's not a normal context menu entry ?

MontaGhanmy:
> image
> 
> it's there and not clickable (my account in that screen shot has no first/last name but email is `a@b.c` ), should we show the e-mail as well ? Is there a way to make it obvious it's not a normal context menu entry ?

@ericlinagora It's not supposed to be clickable (it's text). The email is there is if COMPANY_DISPLAY_EMAIL is enabled (like the way it was before removing the name)
image

shepilov (approved):

PR #733, opened 2024-11-11
🔊 upload logs from frontend

ericlinagora:


tdrive/backend/node/src/services/user/web/controller.ts
```diff
@@ -93,6 +94,30 @@ export class UsersCrudController
...
   }
 
+  /** This allows a logged in user to upload log entries */
+  async reportClientLog(
+    request: FastifyRequest<{ Body: { [key: string]: string } }>,
+  ): Promise<object> {
+    const headers = { ...request.headers };
+    const boringOrSecretHeaders =
+      "cookie authorization cache-control connection pragma content-length content-type accept accept-encoding accept-language".split(
+        /\s+/g,
```
shepilov, line 104: really? split instead of a string list?

tdrive/frontend/src/app/features/users/api/user-api-client.ts
```diff
@@ -142,6 +143,30 @@ class UserAPIClientService {
     });
   }
 
+  /** Send log entry to the server's logs. Try to include a `message` property please */
+  async reportLog(body: any | { message: string, error?: Error, err?: Error, e?: Error, }): Promise<void> {
+    const actualBody = {
+      version: version.version_detail || version.version,
+      ...body,
+    } as any;
+    for (const [key, val] of Object.entries(actualBody)) {
```
shepilov, line 152: why do we need "for" for this? why not just body.err, body.e or whatever, 
Also you are designing a new API, why so complex with multiple errors?

shepilov:
also, could you please make it optional? 
as a feature for the company, and also as a setting for a user

tdrive/frontend/src/app/features/users/api/user-api-client.ts
ericlinagora: it's more convenient, for ex if you're in nested try{}catch(), it's also useful for multiple errors sometimes with different reasons, for ex. an problem that happens when two errors both happen etc. This means there's no convention, nothing to learn and know to use it.

ericlinagora:
> also, could you please make it optional? as a feature for the company, and also as a setting for a user

what does this mean ? if the client doesn't use it, then it's like it doesn't exist. The user for sure shouldn't even know about it. Even if the server doesnt have the route it could still work using nginx logs for ex.

tdrive/backend/node/src/services/user/web/controller.ts
ericlinagora: yes

shepilov: why

tdrive/frontend/src/app/features/users/api/user-api-client.ts
shepilov: you have only on call of this function, and I don't seed nested try catch there. also nested try catch looks like bad design. I don't understand why can two error happens at the same time?

shepilov:
I mean this feature is needed only in "debug" mode, you are putting this "onerror" to all the logs from the client side, and don't know what errors will be looged in our code, and also lib logs as well. I'm not sure that we are writing to err only things that we need to write there.
it would be great to be able to turn off this feature if needed. Or if we are speaking about the future, when the app will be stable, to turn on it only if it's needed, if we know that particular user/company has a problem, and we need to troubleshoot this specific case.
Right now we are writing to console.err things that are not actually errors, specifically during the auth flow.
If you want to do it without feature flag, please, don't write everything, and just call this function where you really you need it.

tdrive/backend/node/src/services/user/web/controller.ts
ericlinagora: I'm not a fan of typing repeating `", "`, it's less likely to take vertical space after the linter, it'll be easier to make configurable if we want to one day, also, why not ?

tdrive/frontend/src/app/features/users/api/user-api-client.ts
ericlinagora: Ah, good point I should splat AggregateErrors, they're annoying to log today. There's a bunch of reasons for multiple errors, say you have 2 apis, try the newer, it fails, try the fallback, it fails, log both. But also, it just means not having to know what keys are magic.

ericlinagora:
I only really care about this in production, in debug I can see client logs just fine. The point of this is to test the variety of browsers and stuff in actual use. If the error comes from a lib I want it as much as any other error. If we do decide it's logging too many errors we don't care about, we can just catch that one, ultimately we should not be having unhandled exceptions - or we should be reloading because state is no longer predictable. If the app is stable, then this feature has 0 cost, if not, it costs a log entry, it's pretty standard practice. I don't use the console, everything uploaded is either an unhandled error, or explicitly requested from the client. I prefer to not make this optional: should we want to, it would be trivial to stop logging server side, but also we shouldn't want to, we should fix our errors or at least be aware of them. Making it optional introduces multiple ways for this to fail without being detected server side (configuration/helm/environment/whatever error), and it encourages turning this off, which is just a head in sand strategy. So because it doesn't get us anything, introduces risk, complicates deployments, I think it's best to keep this simple.

tdrive/backend/node/src/services/user/web/controller.ts
shepilov: support, everybody who reads this code will have this question "why"

tdrive/frontend/src/app/features/users/api/user-api-client.ts
shepilov: there should be two different function calls, one for API and another one for fallback, let's not imagine things and keep it KISS, not also leave in the world where we don't have nested try catches

tdrive/backend/node/src/services/user/web/controller.ts
ericlinagora: I've been doing this trick since the days of ruby, you're the first to ask about it... Actually second

shepilov (approved):

PR #734, opened 2024-11-16
🩹 back: tolerate extraneous `/`s in env vars

ericlinagora:
Prevent further config issues to mitigate internal issue reference 3842

### DIFF
--- a/tdrive/backend/node/src/core/platform/services/email-pusher/index.ts
+++ b/tdrive/backend/node/src/core/platform/services/email-pusher/index.ts
@@ -16,6 +16,7 @@ import path from "path";
 import { existsSync } from "fs";
 import axios from "axios";
 import short, { Translator } from "short-uuid";
+import { joinURL } from "../../../../utils/urls";
 
 export default class EmailPusherClass
   extends TdriveService<EmailPusherAPI>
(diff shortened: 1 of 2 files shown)


tdrive/backend/node/src/core/platform/services/email-pusher/index.ts
```diff
@@ -86,7 +87,14 @@ export default class EmailPusherClass
       const encodedCompanyId = translator.fromUUID(data.notifications[0].item.company_id);
       const encodedItemId = translator.fromUUID(data.notifications[0].item.id);
       const previewType = data.notifications[0].item.is_directory ? "d" : "preview";
-      const encodedUrl = `${this.platformUrl}/client/${encodedCompanyId}/v/shared_with_me/${previewType}/${encodedItemId}`;
+      const encodedUrl = joinURL([
```
shepilov, line 90: this url in unredable, can we have the same syntax as it was before, but just trim "/" in the `platformUrl`?

ericlinagora: again with your opinion on readability......

no, that wouldn't be a generic and safe solution like this is. It should be url encoded too, each component should be trimmed, and it should be the same logic for all url generation. you have an objective reason against this ?

shepilov: previewType = "d" : "preview" why do you need to encode this? the same with encodedCompanyId
and I don't see any reason why it should be generic

ericlinagora: this is standard escaping to prevent injection, same concept as html or script etc anyway, output only correct urls, don't allow injections. This is the 1000th time we've had issues with malformed urls and extra / etc. This is why it should be generic, there is only one correct url format and how to generate it from data should be the same app wide.

shepilov (approved):

PR #736, opened 2024-11-20
⌨️ Keyboard shortcuts/navigation for create and upload

### LINKED ISSUE #224: [Feature] Improve Accessibility
Work on improving the accessibility of the user interface by adding appropriate ARIA roles, alt text for images, and ensuring keyboard navigation is smooth.

MontaGhanmy:
demo_shortcut (online-video-cutter com)

### DIFF
--- a/tdrive/frontend/src/app/atoms/button/button.tsx
+++ b/tdrive/frontend/src/app/atoms/button/button.tsx
@@ -1,6 +1,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import React from 'react';
+import React, { useEffect } from 'react';
 import _ from 'lodash';
+import { addShortcut, removeShortcut } from 'app/features/global/services/shortcut-service';
 
 export type ButtonTheme = 'primary' | 'secondary' | 'danger' | 'default' | 'outline' | 'dark' | 'white' | 'green';
 
@@ -12,6 +13,7 @@ interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   loading?: boolean;
   disabled?: boolean;
   children?: React.ReactNode;
+  shortcut?: string;
 }
 
 export const Button = (props: ButtonProps) => {
@@ -42,9 +44,7 @@ export const Button = (props: ButtonProps) => {
     className =
       'text-zinc-300 border-0 bg-zinc-900 hover:bg-zinc-800 hover:text-white active:bg-zinc-900';
 
-  if (props.theme === 'green')
-    className =
-      'text-zinc-300 border-0 bg-green-700';
+  if (props.theme === 'green') className = 'text-zinc-300 border-0 bg-green-700';
 
   if (disabled) className += ' opacity-50 pointer-events-none';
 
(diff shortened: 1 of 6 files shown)


tdrive/frontend/src/app/atoms/button/button.tsx
```diff
@@ -58,6 +58,46 @@ export const Button = (props: ButtonProps) => {
     else className = className + ' w-9 !p-0 justify-center';
   }
 
+  // translate shortcuts
+  const translateDynamicShortcut = (shortcut: string): string =>
+    shortcut.startsWith('CmdOrCtrl+Shift')
+      ? `⇧⌘${shortcut.replace('CmdOrCtrl+Shift+', '')}`
```
ericlinagora: this symbol means nothing to non mac users =/ i think we should use https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform to show control if it isn't mac

tdrive/frontend/src/app/atoms/button/button.tsx
```diff
@@ -58,6 +58,46 @@ export const Button = (props: ButtonProps) => {
...
+      : shortcut.startsWith('CmdOrCtrl')
+      ? `⌘${shortcut.replace('CmdOrCtrl+', '')}`
+      : shortcut;
+
+  if (props.shortcut) {
+    useEffect(() => {
```
ericlinagora: i think the if (shortcut) should be inside the useEffect callback, because useEffect and friends need to always be called in the same exact order and sequence

tdrive/frontend/src/app/atoms/button/button.tsx
```diff
@@ -58,6 +58,46 @@ export const Button = (props: ButtonProps) => {
...
+        shortcut,
+        handler: event => {
+          handler(event as KeyboardEvent);
+        },
+      });
+
+      return () => {
+        removeShortcut({
+          shortcut,
+          handler: event => {
```
ericlinagora: This function instance will not be the same as the one when added. If the event source tests for the handler it will fail, if it doesn't, this is useless

tdrive/frontend/src/app/atoms/button/button.tsx
```diff
@@ -58,6 +58,46 @@ export const Button = (props: ButtonProps) => {
     else className = className + ' w-9 !p-0 justify-center';
   }
 
+  // translate shortcuts
+  const translateDynamicShortcut = (shortcut: string): string =>
```
ericlinagora: This function should probably be in `app/features/global/services/shortcut-service`

tdrive/frontend/src/app/atoms/button/button.tsx
```diff
@@ -58,6 +58,46 @@ export const Button = (props: ButtonProps) => {
...
+      const handler = (event?: KeyboardEvent) => {
+        // Disable default behavior for the shortcut
+        if (event) {
+          event.preventDefault();
+          event.stopPropagation();
+        }
+        props.onClick && props.onClick({} as any);
+      };
+
+      const shortcut = props.shortcut || '';
```
ericlinagora: this is inside `if (props.shortcut)` so it's always truthy here

tdrive/frontend/src/app/views/client/body/drive/modals/create/index.tsx
```diff
@@ -168,10 +168,19 @@ export const CreateModal = ({
...
+      props.onClick();
+    }
+  };
+
   return (
     <div
       onClick={props.onClick}
-      className="flex flex-row p-4 dark:bg-zinc-900 dark:text-white bg-zinc-100 hover:bg-opacity-75 cursor-pointer rounded-md m-2"
+      className="flex flex-row p-4 dark:bg-zinc-900 dark:text-white bg-zinc-100 hover:bg-opacity-75 cursor-pointer rounded-md m-2 focus:bg-zinc-800 dark:focus:bg-zinc-800 outline-none focus:border-none"
+      tabIndex={0}
```
ericlinagora, line 187: why the `{}` and not just 0 ? is this required at all, we're not really doing the tab indexing stuff anywhere else ?

tdrive/frontend/src/app/views/client/side-bar/actions.tsx
```diff
@@ -162,6 +162,7 @@ export default () => {
 
               <Button
                 onClick={() => uploadItemModal()}
+                shortcut='CmdOrCtrl+U'
```
ericlinagora: is the modifier necessary ?

MontaGhanmy: Yes

tdrive/frontend/src/app/views/client/body/drive/modals/create/index.tsx
MontaGhanmy: tabIndex is a number that's why. It is required for tab to work/navigate the modal items.

ericlinagora: html dom attribute values are always strings anyway, see https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute . what's happening here is just making react type it first and maybe not guess its a constant it could optimise out

ericlinagora (approved):

PR #738, opened 2024-11-20
🐛 Fix email notification link

MontaGhanmy:
demo_scroll

### DIFF
--- a/tdrive/backend/node/src/core/platform/services/email-pusher/index.ts
+++ b/tdrive/backend/node/src/core/platform/services/email-pusher/index.ts
@@ -42,6 +42,7 @@ export default class EmailPusherClass
     });
     this.interface = this.configuration.get<string>("email_interface", "");
     this.platformUrl = this.configuration.get<string>("platform_url", "");
+    this.debug = this.configuration.get<boolean>("debug", false);
     if (this.interface === "smtp") {
       const useTLS = this.configuration.get<string>("smtp_tls", "false") == "true";
       const smtpConfig: SMTPClientConfigType = {
@@ -59,7 +60,6 @@ export default class EmailPusherClass
       this.apiUrl = this.configuration.get<string>("endpoint", "");
       this.apiKey = this.configuration.get<string>("api_key", "");
       this.sender = this.configuration.get<string>("sender", "");
-      this.debug = this.configuration.get<boolean>("debug", false);
     }
 
     return this;
@@ -107,7 +107,7 @@ export default class EmailPusherClass
 
       return { html, text, subject };
     } catch (error) {
-      this.logger.error(`Failure when building email template: ${error}`);
+      this.logger.error({ error }, `Failure when building email template: ${error}`);
     }
   }
 
@@ -139,7 +139,7 @@ export default class EmailPusherClass
       };
 
       if (this.debug) {
-        this.logger.info("EMAIL::SENT ", { emailObject });
+        this.logger.info({ emailObject }, "EMAIL::SENT");
       } else {
         if (this.interface === "smtp") {
           try {
--- a/tdrive/backend/node/src/core/platform/services/email-pusher/index.ts
+++ b/tdrive/backend/node/src/core/platform/services/email-pusher/index.ts
@@ -15,7 +15,6 @@ import { convert } from "html-to-text";
 import path from "path";
 import { existsSync } from "fs";
 import axios from "axios";
-import short, { Translator } from "short-uuid";
 import { joinURL } from "../../../../utils/urls";
 
 export default class EmailPusherClass
@@ -81,20 +80,10 @@ export default class EmailPusherClass
   ): Promise<EmailBuilderRenderedResult> {
     try {
       language = ["en", "fr"].find(l => language.toLocaleLowerCase().includes(l)) || "en";
-      const translator: Translator = short();
       const templatePath = path.join(__dirname, "templates", language, `${template}.eta`);
       const subjectPath = path.join(__dirname, "templates", language, `${template}.subject.eta`);
-      const encodedCompanyId = translator.fromUUID(data.notifications[0].item.company_id);
-      const encodedItemId = translator.fromUUID(data.notifications[0].item.id);
-      const previewType = data.notifications[0].item.is_directory ? "d" : "preview";
-      const encodedUrl = joinURL([
-        this.platformUrl,
-        "client",
-        encodedCompanyId,
-        "v/shared_with_me",
-        previewType,
-        encodedItemId,
-      ]);
+      const urlComponents = data.notifications[0].urlComponents;
+      const encodedUrl = joinURL([this.platformUrl, ...urlComponents]);
 
       if (!existsSync(templatePath)) {
         throw Error(`template not found: ${templatePath}`);
--- a/tdrive/backend/node/src/services/documents/services/engine/index.ts
+++ b/tdrive/backend/node/src/services/documents/services/engine/index.ts
@@ -7,10 +7,14 @@ import { DocumentsProcessor } from "./extract-keywords";
 import Repository from "../../../../core/platform/services/database/services/orm/repository/repository";
 import { DriveFile, TYPE } from "../../entities/drive-file";
 import { DocumentsFinishedProcess } from "./save-keywords";
+import short, { Translator } from "short-uuid";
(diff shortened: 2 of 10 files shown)


tdrive/backend/node/src/services/documents/services/engine/index.ts
```diff
@@ -2,15 +2,25 @@ import globalResolver from "../../../global-resolver";
...
 import { DocumentsProcessor } from "./extract-keywords";
 import Repository from "../../../../core/platform/services/database/services/orm/repository/repository";
 import { DriveFile, TYPE } from "../../entities/drive-file";
 import { DocumentsFinishedProcess } from "./save-keywords";
+import short, { Translator } from "short-uuid";
+import { getConfigOrDefault } from "../../../../utils/get-config";
+import fs from "fs";
 export class DocumentsEngine implements Initializable {
   private documentRepository: Repository<DriveFile>;
+  private platformUrl: string = getConfigOrDefault("drive.defaultUserQuota", 0);
```
shepilov: platformUrl - userQuota? 
@MontaGhanmy could you please write unit tests for the function, that returns file URL.

tdrive/backend/node/src/services/documents/services/engine/index.ts
```diff
@@ -25,6 +35,33 @@ export class DocumentsEngine implements Initializable {
       return; // Early return on unknown event type
     }
 
+    const encodedCompanyId = translator.fromUUID(e.item.company_id);
```
shepilov: Extract this part to a one function with two params, item or item path, and receiver. And have the next test cases:
- file created in the folder that I've shared and the folder Is in: my drive, shared drive
- the file was shared with me
- any other cases that I forgot)

I home we will refactor it later, when we refactor UI routings wot have one file URL

tdrive/backend/node/src/services/documents/types.ts
```diff
@@ -125,9 +125,15 @@ export const eventToTemplateMap: Record<string, any> = {
   [DocumentEvents.DOCUMENT_SAHRED]: "notification-document-shared",
 };
 
+export enum NotificationActionType {
```
shepilov, line 128: and what does it mean, "direct" or "update" ?

MontaGhanmy: - "direct" is a user A share folder/file with user B.
- "update" is user B does an update on a file/folder shared by user A. (this could be also an anonymous user through public links)

tdrive/backend/node/src/core/platform/services/email-pusher/index.ts
```diff
@@ -81,20 +80,10 @@ export default class EmailPusherClass
...
-      const previewType = data.notifications[0].item.is_directory ? "d" : "preview";
-      const encodedUrl = joinURL([
-        this.platformUrl,
-        "client",
-        encodedCompanyId,
-        "v/shared_with_me",
-        previewType,
-        encodedItemId,
-      ]);
+      const urlComponents = data.notifications[0].urlComponents;
```
ericlinagora, line 85: I don't entirely understand, if there is more than one notification, the link should be specific to each notification, so there should be many no ?

MontaGhanmy: There is no more than one notifications, legacy code used to do it that way, the email pusher needs to be refactored but that's beyond the scope of this PR.

tdrive/backend/node/src/services/documents/utils.ts
```diff
@@ -673,3 +676,38 @@ export const getKeywordsOfFile = async (
...
+  const translator: Translator = short();
+  const encodedCompanyId = translator.fromUUID(e.item.company_id);
+  const clientPath = ["client", encodedCompanyId, "v"];
+  const isPersonalScope = e.item.scope === "personal";
+  const isDirectory = e.item.is_directory;
+  const itemId = isDirectory ? e.item.id : e.item.parent_id;
+
+  // Determine the scope
+  let view;
+  if (e.type === NotificationActionType.UPDATE) {
```
ericlinagora: with enums, as a matter of good practice, its important to fail for unexpected values, so like
```
switch(enum) {
case x: .... break;
case y: ... break;
default throw new Error(`unexpected NotificationActionType value ${JSON.stringify(e.type)}`);
}
```

tdrive/backend/node/src/services/documents/utils.ts
```diff
@@ -673,3 +676,38 @@ export const getKeywordsOfFile = async (
...
+    view = isPersonalScope ? `user_${receiver}` : "root";
+  } else {
+    view = "shared_with_me";
+  }
+
+  // Build URL components
+  const urlComponents = [...clientPath, view];
+
+  // Add directory and itemId if applicable
+  if (e.type === "update" || isDirectory) {
```
ericlinagora: `NotificationActionType.UPDATE`

tdrive/backend/node/test/unit/core/services/documents/document-notification-url-components.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    scope: "personal",
+    is_directory: false,
+    is_in_trash: false,
+    ...overrides,
+  });
+
+  const mockPayload = (
+    type: NotificationActionType,
+    overrides: Partial<NotificationPayloadType> = {},
+  ) => ({
```
ericlinagora, line 32: `: NotificationPayloadType` this way the `as`s bellow aren't needed

tdrive/frontend/src/app/features/router/services/router-service.ts
```diff
@@ -54,7 +54,7 @@ class RouterServices extends Observable {
     '/client/:companyId/v/:viewId',
     '/client/:companyId/v/:viewId/d/:dirId',
     '/client/:companyId/v/:viewId/preview/:itemId',
-    '/client/:companyId/v/:viewId/d/:dirId/preview/:itemId',
+    '/client/:companyId/v/:viewId/d/:dirId/preview/:itemId', // preview item
```
ericlinagora: it's in the url tho

tdrive/frontend/src/app/views/client/body/drive/browser.tsx
```diff
@@ -107,7 +107,7 @@ export default memo(
       children,
       loading: loadingParent,
       path,
-      loadNextPage,
+      loadNextPage
```
ericlinagora: in general i think we should leave the trailing commas. Not really important, but if we add a param, then we have 2 lines that changed

tdrive/frontend/src/app/views/client/body/drive/browser.tsx
```diff
@@ -259,6 +259,25 @@ export default memo(
...
+    useEffect(() => {
+      const itemInChildren = children.find(item => item.id === scrollItemId);
+      if (!loading && scrollTillItemInView && !itemInChildren) {
+        scrollViwer.current?.scrollTo(0, scrollViwer.current?.scrollHeight);
+      } else {
+        if (!loading && itemInChildren) {
+          // scroll to id using id
+          const element = document.getElementById(scrollItemId);
+          element?.scrollIntoView({ behavior: 'smooth', block: 'center' });
+          // set as checked
```
ericlinagora: these comments aren't really adding much...

tdrive/frontend/src/app/views/client/body/drive/documents/document-row.tsx
```diff
@@ -54,6 +54,7 @@ export const DocumentRow = ({
           : 'hover:bg-zinc-500 hover:bg-opacity-10 ') +
         (className || '')
       }
+      id={item.id}
```
ericlinagora: Can we either use another attribute (eg: `data-item-id={item.id}` with `querySelector`), or if we really want `id`, then prefix it in case we have multiple items on screen representing the same driveitem

tdrive/backend/node/test/unit/core/services/documents/document-notification-url-components.test.ts
MontaGhanmy: That would require specifying all the properties of NotificationPayloadType even for test cases where they're irrelevant. It makes it even more complicated.

ericlinagora (approved):

PR #760, opened 2024-12-03
✨ Load pages and ensure scrollbar is always visible

### LINKED ISSUE #755: When the first page of items loads without needing a scrollbar, no way to get to second page
Depending on window size, font size and page count, if the first page loads but doesn't have enough height to trigger the scroll bar to show, then there is never the event to start loading the second page.

If possible, keep loading pages while the scrollbar isn't visible yet. And maybe increase default page item count. After first load, use height of an item and the window to load a known good number ?

Note: update ticker 3844 in internal system too.

MontaGhanmy:


ericlinagora (approved):

ericlinagora:
I'm just a bit curious about some more esoteric scenarios: you load a first page, it has a scrollbar, font too big you zoom out, no more scroll bar =D I think we can let it be since a refresh fixes it

MontaGhanmy:
> I'm just a bit curious about some more esoteric scenarios: you load a first page, it has a scrollbar, font too big you zoom out, no more scroll bar =D I think we can let it be since a refresh fixes it

@ericlinagora I just updated it so it does try to ensure the scroll bar is in view when the window is re-sized;

tdrive/frontend/src/app/views/client/body/drive/browser.tsx
```diff
@@ -241,32 +241,33 @@ export default memo(
...
+      const clientHeight = scrollViewer.current?.clientHeight || 0;
       if (scrollTop > 0 && scrollTop + clientHeight >= scrollHeight) {
         await loadNextPage(parentId);
       }
     };
 
     useEffect(() => {
-      if(!loading) scrollViwer.current?.addEventListener('scroll', handleScroll, { passive: true });
+      if (!loading)
+        scrollViewer.current?.addEventListener('scroll', handleScroll, { passive: true });
```
lethemanh, line 257: Everything looks good to me, but have you considered replacing the scroll event with the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) in the future? I saw that the scroll event triggers even when the scrollbar hasn’t reached the target point.

ericlinagora: I didn't know about that API, looks great thanks

PR #763, opened 2024-12-08
probe and diagnostic endpoints

### LINKED ISSUE #762: Probe and diagnostic endpoints
Implement a pluggable but reliable way to introspect the node process state from within, as well as detect critical failures.

Probing endpoints:

- require the query param: `?secret=value-of-DIAG_PROBE_SECRET`
- `/api/diagnostics/t/startup`, `/api/diagnostics/t/ready`, `/api/diagnostics/t/live`: correspond to the respective kubernetes probes
- `/api/diagnostics/t/stats`: more exhaustive statistics endpoint

Environment variables:
- `DIAG_PROBE_SECRET` secret passed as query parameter to all endpoints
- `DIAG_SECRET` secret never passed over network or in logs, required for more advanced debugging functions

ericlinagora:
basic lifecycle probes for kubes and an endpoint for diagnosing mem leaks

missing tests for av, message queue, possibly sso

### DIFF
--- a/tdrive/backend/node/config/custom-environment-variables.json
+++ b/tdrive/backend/node/config/custom-environment-variables.json
@@ -21,7 +21,9 @@
       "__name": "DIAG_SKIP_KEYS",
       "__format": "json"
     },
-    "probeSecret": "DIAG_PROBE_SECRET"
+    "probeSecret": "DIAG_PROBE_SECRET",
+    "statsLogPeriodMs": "DIAG_STATS_PRINT_PERIOD_MS",
+    "statsFullStatsLogPeriodMs": "DIAG_FULL_STATS_PRINT_PERIOD_MS"
   },
   "webserver": {
     "host": "TWAKE_DRIVE_HOST",
--- a/tdrive/backend/node/config/custom-environment-variables.json
+++ b/tdrive/backend/node/config/custom-environment-variables.json
@@ -16,6 +16,14 @@
   "logger": {
     "level": "LOG_LEVEL"
   },
+  "diagnostics": {
+    "skipKeys": {
+      "__name": "DIAG_SKIP_KEYS",
+      "__format": "json"
+    },
+    "probeSecret": "DIAG_PROBE_SECRET",
+    "secret": "DIAG_SECRET"
+  },
   "webserver": {
     "host": "TWAKE_DRIVE_HOST",
     "logger": {
--- a/tdrive/backend/node/config/custom-environment-variables.json
+++ b/tdrive/backend/node/config/custom-environment-variables.json
@@ -21,8 +21,7 @@
       "__name": "DIAG_SKIP_KEYS",
       "__format": "json"
     },
-    "probeSecret": "DIAG_PROBE_SECRET",
-    "secret": "DIAG_SECRET"
+    "probeSecret": "DIAG_PROBE_SECRET"
   },
   "webserver": {
     "host": "TWAKE_DRIVE_HOST",
--- a/tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/postgres/postgres.ts
+++ b/tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/postgres/postgres.ts
@@ -74,10 +74,24 @@ export class PostgresConnector extends AbstractConnector<PostgresConnectionOptio
     switch (depth) {
       case TServiceDiagnosticDepth.alive:
         return { ok: true, didConnect: await this.ping() };
-      case TServiceDiagnosticDepth.stats_basic:
       case TServiceDiagnosticDepth.stats_track:
+        return {
+          ok: true,
+          db: (
+            await this.client.query("select * from pg_stat_database where datname = $1", [
+              this.options.database,
+            ])
+          ).rows,
+        };
+      case TServiceDiagnosticDepth.stats_basic:
+        return { ok: true, warn: "pgsql_basic_has_nothing_more_than_track" };
       case TServiceDiagnosticDepth.stats_deep:
-        return { ok: true, warn: "unsupported_depth" };
+        return {
+          ok: true,
+          databases: (await this.client.query("select * from pg_stat_database")).rows,
+          tables: (await this.client.query("select * from pg_stat_user_tables")).rows,
+          indexes: (await this.client.query("select * from pg_stat_user_indexes")).rows,
+        };
 
       default:
         throw new Error(`Unexpected TServiceDiagnosticDepth: ${JSON.stringify(depth)}`);
--- a/tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/postgres/postgres.ts
+++ b/tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/postgres/postgres.ts
@@ -71,26 +71,41 @@ export class PostgresConnector extends AbstractConnector<PostgresConnectionOptio
   }
 
   async getDiagnostics(depth: TServiceDiagnosticDepth): Promise<TDiagnosticResult> {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const safeRequest = async (query: string, values?: any[]) => {
+      try {
+        return (await this.client.query(query, values)).rows;
+      } catch (err) {
+        const logId = "pg-diags-error-" + Math.floor(process.uptime() * 1000);
+        logger.error(
+          { err, query, values, logId, errCode: err.code },
+          `Error running postgresql statistics at ${depth} ( ${logId} ) `,
+        );
+        return { error: true, logId };
+      }
+    };
     switch (depth) {
+      // This is the only required `ok`
       case TServiceDiagnosticDepth.alive:
         return { ok: true, didConnect: await this.ping() };
+
+      // Statistics can silently fail, and do it granularly if there is
+      // a permission issue only on some of the stats
       case TServiceDiagnosticDepth.stats_track:
         return {
           ok: true,
-          db: (
-            await this.client.query("select * from pg_stat_database where datname = $1", [
-              this.options.database,
-            ])
-          ).rows,
+          db: await safeRequest("select * from pg_stat_database where datname = $1", [
+            this.options.database,
+          ]),
         };
       case TServiceDiagnosticDepth.stats_basic:
-        return { ok: true, warn: "pgsql_basic_has_nothing_more_than_track" };
+        return { ok: true, warn: "pgsql_basic_has_no_basic_level_stats" };
       case TServiceDiagnosticDepth.stats_deep:
         return {
           ok: true,
-          databases: (await this.client.query("select * from pg_stat_database")).rows,
-          tables: (await this.client.query("select * from pg_stat_user_tables")).rows,
-          indexes: (await this.client.query("select * from pg_stat_user_indexes")).rows,
+          databases: await safeRequest("select * from pg_stat_database"),
+          tables: await safeRequest("select * from pg_stat_user_tables"),
+          indexes: await safeRequest("select * from pg_stat_user_indexes"),
         };
 
       default:
--- a/tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/postgres/postgres.ts
+++ b/tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/postgres/postgres.ts
@@ -8,6 +8,10 @@ import { getEntityDefinition } from "../../../orm/utils";
 import { UpsertOptions } from "src/core/platform/services/database/services/orm/connectors";
 import { PostgresDataTransformer, TypeMappings } from "./postgres-data-transform";
 import { PostgresQueryBuilder, Query } from "./postgres-query-builder";
+import {
+  TDiagnosticResult,
+  TServiceDiagnosticDepth,
+} from "../../../../../../framework/api/diagnostics";
 
 export interface PostgresConnectionOptions {
   database: string;
@@ -25,6 +29,7 @@ export interface PostgresConnectionOptions {
 export class PostgresConnector extends AbstractConnector<PostgresConnectionOptions> {
   private logger = getLogger("PostgresConnector");
   private client: Client = new Client(this.options);
+  private connected = false;
   private dataTransformer = new PostgresDataTransformer({ secret: this.secret });
   private queryBuilder = new PostgresQueryBuilder(this.secret);
 
@@ -40,10 +45,12 @@ export class PostgresConnector extends AbstractConnector<PostgresConnectionOptio
     if (this.client) {
       this.logger.info("Connecting to DB");
       this.client.on("error", err => {
-        this.logger.warn(err, "PostgreSQL connection error");
+        this.logger.error(err, "PostgreSQL connection error");
       });
+      this.client.on("end", () => (this.connected = false));
 
       await this.client.connect();
+      this.connected = true;
       this.logger.info("Connection pool created");
       await this.healthcheck();
     }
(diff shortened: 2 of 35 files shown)


tdrive/backend/node/src/core/platform/services/diagnostics/utils.ts
```diff
@@ -0,0 +1,40 @@
...
+  return { durationMs: new Date().getTime() - startMs, result };
+}
+
+/** Use `node:inspector` to create a snapshot of the heap, and
+ * synchroneously (because of session.post) pipe it to a stream
+ * readable from the first argument of the callback.
+ */
+export function getHeapSnapshotSync(cb: (readable: Readable) => void) {
+  logger.info({ gcExposed: !!global.gc }, "Beginning heap snapshot");
+  if (global.gc) global.gc();
```
shepilov, line 20: why are you running gc if you need only statistics?

tdrive/backend/node/config/custom-environment-variables.json
```diff
@@ -16,6 +16,14 @@
   "logger": {
     "level": "LOG_LEVEL"
   },
+  "diagnostics": {
+    "skipKeys": {
+      "__name": "DIAG_SKIP_KEYS",
+      "__format": "json"
+    },
+    "probeSecret": "DIAG_PROBE_SECRET",
```
shepilov, line 24: why do we need to secrets?

and why not to user application authentication for this?

tdrive/backend/node/src/core/platform/services/storage/oneof-storage-strategy.ts
```diff
@@ -34,6 +35,17 @@ export class OneOfStorageStrategy implements StorageConnectorAPI {
     return this.id;
   }
 
+  async getDiagnostics(depth: TServiceDiagnosticDepth): Promise<TDiagnosticResult> {
+    const states = await Promise.all(
+      this.storages.map(async s => ({ id: s.getId(), ...(await s.getDiagnostics(depth)) })),
```
shepilov, line 40: could you please add id of the storage, at this level during map or inside every storage statistics?

tdrive/backend/node/src/core/platform/services/diagnostics/providers/process.ts
```diff
@@ -0,0 +1,16 @@
+import diagnostics from "../../../framework/api/diagnostics";
+
+export default () =>
+  diagnostics.registerProviders({
+    key: "process",
+    tags: ["live", "ready"],
+    async get() {
+      return {
+        ok: true,
+        gc: !!global.gc,
```
shepilov, line 15: what this information gives you?

tdrive/backend/node/src/core/platform/services/database/services/orm/connectors/postgres/postgres.ts
```diff
@@ -56,6 +63,41 @@ export class PostgresConnector extends AbstractConnector<PostgresConnectionOptio
...
+              this.options.database,
+            ])
+          ).rows,
+        };
+      case TServiceDiagnosticDepth.stats_basic:
+        return { ok: true, warn: "pgsql_basic_has_nothing_more_than_track" };
+      case TServiceDiagnosticDepth.stats_deep:
+        return {
+          ok: true,
+          databases: (await this.client.query("select * from pg_stat_database")).rows,
```
shepilov: If you are running this query not from database owner, your user should have permissions to pg_stat_database, if it's possible it would be great to catch this exception here, and translate it explicitly to clients error

tdrive/backend/node/test/e2e/diagnostics/diagnostics.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    platform = null;
+  });
+
+  const getDiagnosticTags = (tag: string, secret: string) =>
+    platform.app.inject({
+      method: "GET",
+      url: `/api/diagnostics/t/${encodeURIComponent(tag)}?secret=${encodeURIComponent(secret)}`,
+    });
+
+  it("should refuse invalid probe secrets", async () => {
```
shepilov, line 42: add ok tests please with checks for storage, database, process statistics

tdrive/backend/node/src/core/platform/services/diagnostics/web/probe-routes.ts
```diff
@@ -0,0 +1,25 @@
+import type { FastifyInstance, FastifyPluginCallback } from "fastify";
+import diagnostics, {
+  getConfig as getDiagnosticsGetConfig,
+} from "../../../framework/api/diagnostics";
+
+const routes: FastifyPluginCallback = (fastify: FastifyInstance, _opts, next) => {
+  const diagnosticsConfig = getDiagnosticsGetConfig();
+  if (diagnosticsConfig?.probeSecret?.length) {
```
shepilov, line 8: it's a feature flag?)

tdrive/backend/node/src/core/platform/services/diagnostics/web/probe-routes.ts
```diff
@@ -0,0 +1,25 @@
...
+    const tagParamName = "diagnosticTags";
+    fastify.get(`/t/:${tagParamName}`, async (request, reply) => {
+      const tag = request.params[tagParamName];
```
shepilov, line 11: why do er need to pass param name in URL?

tdrive/backend/node/src/core/platform/services/diagnostics/utils.ts
ericlinagora: this isn't for statistics, it's a heap snapshot, so i want it with as little stuff left that would be soon be collected.

tdrive/backend/node/config/custom-environment-variables.json
ericlinagora: the secrets are documented in the config interface. One needs to be very minimal and simple and state less for kube to auth itself, this ensure basic things like it's a bit harder to get stats from the public, but, devs still can. kube isn't going through the app auth process. Neither is a dev trying to get a snapshot to diagnose mem leaks in situ, but that needs tougher security, yet accessible from a simple script

tdrive/backend/node/src/core/platform/services/diagnostics/providers/process.ts
ericlinagora: whether node was started with the `--expose-gc` flag, which would be best for memleak diagnostic

ericlinagora: (and if not, to treat the snapshots as having potentially uncollected but not leaking allocs)

tdrive/backend/node/src/core/platform/services/diagnostics/web/heap-routes.ts
```diff
@@ -0,0 +1,86 @@
...
+    })();
+    const hashToken = (token: string = getRunningToken()) =>
+      createHash("sha512").update(`+${token}+${diagnosticsConfig.secret}+`).digest("hex");
+
+    /*
+      Example flow:
+        Configuration:
+          { secret: 'secretValue', probeSecret: 'probeSecretValue' }
+        GET current token using the probeSecret:
+          $ curl "$SERVER/api/diagnostics/heap?secret=probeSecretValue"
```
shepilov: we can't forward requests on the load balancer depending on the token, so there is a probability that you get token from one instance and use it on another.

shepilov: why not just pass one secret in the request body? or use oidc auth then with a specific scope

tdrive/backend/node/test/e2e/diagnostics/diagnostics.spec.ts
ericlinagora: storage and db i'd just mock the function on the services that return the implementation's result, so it's not testing anything more than doing the same for the process ok, so just process is ok ?

tdrive/backend/node/src/core/platform/services/diagnostics/web/probe-routes.ts
ericlinagora: Kind of, no secret, no stats routes then. This will help distinguish from a secret error, and fails it safe and light. We could add a separate flag for it, but it just seems like more complexity for no new feature, and also we don't want it active without a secret

tdrive/backend/node/src/core/platform/services/diagnostics/web/probe-routes.ts
ericlinagora: if the question is why did i use a constant it's simply because that name doesn't matter anymore than being the same in those two places

tdrive/backend/node/src/core/platform/services/diagnostics/web/heap-routes.ts
ericlinagora: Correct, but the proper solution to that is store the token in something shared, or to not use one for the hash entropy etc etc... I'm betting that the call from GET to POST is very quick (there's nearly nothing in the GET), the POST will fail real quick with the wrong token too (only in mem state), so just keep retrying the POST. It's a bit of a monte carlo algorithm situation, but given we don't have that many backend instances, i think it's fine. If we get to the point where this is an issue, we can either increase the token timeout (but I wanted that to be hard because it worsens security, so it's ±hardcoded), actually do something with the loadbalancer (it must maintain a list of live nodes anyway), or just figure if we really got that many, do we really care about investigating one of them for leaks ?

tdrive/backend/node/config/custom-environment-variables.json
ericlinagora: (another consideration against app auth, is that one purpose of this is to detect when stuff fails, and auth requires a lot to work already)

tdrive/backend/node/src/core/platform/services/diagnostics/web/heap-routes.ts
ericlinagora: This needs to work hopefully on a nearly dead node, it should be as minimal as possible, and this method prevents the secret from going over the net, through the logs, etc.

tdrive/backend/node/src/core/platform/services/storage/oneof-storage-strategy.ts
ericlinagora: I'm not sure I understood, because that is adding the id ? Want the id at the lower lever from each connector ?

tdrive/backend/node/src/core/platform/services/diagnostics/utils.ts
shepilov: but a picture can be different, and you don't wait, so probably you will take a snapshot during garbage collection

tdrive/backend/node/src/core/platform/services/storage/oneof-storage-strategy.ts
shepilov: if one of the connectors is down, we will understand which one without additional logs search

tdrive/backend/node/src/core/platform/services/diagnostics/providers/process.ts
shepilov: did you add this flag to our docker compose?

tdrive/backend/node/test/e2e/diagnostics/diagnostics.spec.ts
shepilov: yes, we have e2e test running with storage and db, do not mock them

shepilov: if you need to test some specific functionality with mocked services, you can write small init test for this, you don't need all platform started
but we need e2e test for diagnostics as well

tdrive/backend/node/src/core/platform/services/diagnostics/web/probe-routes.ts
shepilov: still don't understand, why you need URL like /t/:diagnosticTags?diagnosticTags=process
why the name of the endpoint "diagnosticTags? ? why do you need t then?

tdrive/backend/node/src/core/platform/services/diagnostics/web/heap-routes.ts
shepilov: ok, that make sense, because we always need to do this call on a specific node instance, we don't have any specific rules on the loadbalancer to route requests, and heap endpoint will be called directly one of the node instance. 
But then we don't need this secret rotation, we can have just only on is the properties as far as we don't expose this URL outside.
And why do you need this casino of the request? If you need to have a heapdump of a specific pod in production(all pods can't have memory problems at the same time), and we have 4 pods in production, if you use a public URL you need 8 requests in average to get heap from the pod you need. And meanwhile you won't understand what pod you need from the first request.
They will be matched in liveness probe in grafan, but this heap is useful only when you sending it to the pod you know. 

I think we should make this URL private, do not expose them and make accessibe in 4000 port, and protect them with one secret in configuraiont.

tdrive/backend/node/src/core/platform/services/diagnostics/providers/process.ts
ericlinagora: Locally for the moment, I was hesitating about including also in the production one, I don't think I can find a downside so far.... I want to see if we can use NODE_OPTIONS too to make it configurable. Ultimately, I don't think this will matter that much in the results, it's really mostly a "best effort" approach

tdrive/backend/node/src/core/platform/services/diagnostics/utils.ts
ericlinagora: Well it's not like JS would otherwise be forgetting to GC, so it is a different picture, but nothing lost is something we don't want to lose. There is no wait on the GC in js ( https://github.com/nodejs/node/blob/5421e15bdc6122420d90ec462c3622b57d59c9d6/deps/v8/src/extensions/gc-extension.cc#L76 ), now I'm pretty sure it's a sync call anyway, but even if it isn't, then getting a dump mid-collect is safe and fine too (this isn't java), and the best we can do is use this endpoint to trigger one, then a bit later call again. Ultimately this is just not much work, probably gives better results, not much more than that.

tdrive/backend/node/src/core/platform/services/diagnostics/web/heap-routes.ts
ericlinagora: yes but prod is the only environment we'd have that problem in, for all others given that months in we still have very limited access, I would like this to be available publicly so devs, the only people who can use this, have access. Using this in prod would be a very last resort, and a dangerous one given it leaks information. We'll always be limited by how much access devs get, if we can get kube access to all other envs, then yeah I'd agree on keeping this internal

tdrive/backend/node/src/core/platform/services/diagnostics/web/probe-routes.ts
ericlinagora: This matches urls like `/t/started` or `/t/ready` etc... I shouldn't have made "tags" plural. The `/t/` is because that's the tag endpoint, to avoid collisions if the tag names evolve. There is already the heap next to t, and we can probably think of other things to add as diagnostics down the road and this would be the spot

tdrive/backend/node/src/core/platform/services/storage/oneof-storage-strategy.ts
ericlinagora: Yes that's what `({ id: s.getId(), ...(await` does, it adds the id but only in the OneOf . I don't understand

tdrive/backend/node/src/core/platform/services/diagnostics/web/heap-routes.ts
shepilov: kube access is another problem and it's partially, solved, so let's make it internal

tdrive/backend/node/src/core/platform/services/storage/oneof-storage-strategy.ts
shepilov: I missed this

shepilov (approved):

PR #764, opened 2024-12-09
Add testing identifier to elements

lethemanh:
Add testing identifier to elements

## How Has This Been Tested?
The test id class names have prefix `testid:`

## Types of changes

- [x] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/tdrive/frontend/src/app/atoms/button/button.tsx
+++ b/tdrive/frontend/src/app/atoms/button/button.tsx
@@ -59,6 +59,8 @@ export const Button = (props: ButtonProps) => {
     else className = className + ' w-9 !p-0 justify-center';
   }
 
+  const testId = props.testClassId ? `testid:${props.testClassId}` : '';
+
   return (
     <button
       type="button"
@@ -68,7 +70,7 @@ export const Button = (props: ButtonProps) => {
         ' ' +
         props.className +
         ' ' +
-        `testid:${props.testClassId}`
+        testId
       }
       disabled={disabled}
       {..._.omit(props, 'loading', 'children', 'className', 'icon', 'iconSize', 'testClassId')}
--- a/tdrive/frontend/src/app/atoms/button/button.tsx
+++ b/tdrive/frontend/src/app/atoms/button/button.tsx
@@ -12,6 +12,7 @@ interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   loading?: boolean;
   disabled?: boolean;
   children?: React.ReactNode;
+  testClassId?: string;
 }
 
 export const Button = (props: ButtonProps) => {
@@ -65,10 +66,12 @@ export const Button = (props: ButtonProps) => {
         ' inline-flex items-center px-4 py-2 border font-medium rounded-md focus:outline-none ' +
         className +
         ' ' +
-        props.className
+        props.className +
+        ' ' +
+        `testid:${props.testClassId}`
       }
       disabled={disabled}
-      {..._.omit(props, 'loading', 'children', 'className', 'icon', 'iconSize')}
+      {..._.omit(props, 'loading', 'children', 'className', 'icon', 'iconSize', 'testClassId')}
     >
       {props.loading && (
         <>
--- a/tdrive/frontend/src/app/atoms/input/input-checkbox-slider.tsx
+++ b/tdrive/frontend/src/app/atoms/input/input-checkbox-slider.tsx
@@ -1,9 +1,11 @@
+import _ from "lodash";
 import { isUndefined } from "lodash";
 
 interface CheckboxSliderProps extends React.InputHTMLAttributes<HTMLInputElement> {
   checked?: boolean;
   disabled?: boolean;
   noFocusBorder?: boolean;
+  testClassId?: string;
 }
 /** Just classes up a checkbox to look like a slider. Deviates from onChange etc for
  * compatibility with `InputHTMLAttributes<HTMLInputElement>`, to expose eg. id to use with htmlFor.
@@ -15,13 +17,17 @@ export const CheckboxSlider = (props: CheckboxSliderProps) => {
 
   return (
     <input
-      {...props}
+      {..._.omit(props, 'testClassId')}
       type="checkbox"
       role="switch"
       style={styles}
       // @author https://tw-elements.com/docs/react/forms/switch/
-      className="!bg-none mr-2 mt-[0.3rem] h-7 w-14 appearance-none rounded-[0.9rem] bg-neutral-300 before:pointer-events-none before:absolute before:h-7 before:w-5 before:rounded-full before:bg-transparent before:content-[''] after:absolute after:z-[2] after:mt-[0.2rem] after:ml-[0.2rem] after:h-5 after:w-5 after:rounded-full after:border-none after:bg-neutral-100 after:shadow-[0_0px_3px_0_rgb(0_0_0_/_7%),_0_2px_2px_0_rgb(0_0_0_/_4%)] after:transition-[background-color_0.2s,transform_0.2s] after:content-[''] checked:bg-primary checked:after:absolute checked:after:z-[2] checked:after:ml-[1.9625rem] checked:after:w-5 checked:after:rounded-full checked:after:border-none checked:after:bg-primary checked:after:shadow-[0_3px_1px_-2px_rgba(0,0,0,0.2),_0_2px_2px_0_rgba(0,0,0,0.14),_0_1px_5px_0_rgba(0,0,0,0.12)] checked:after:transition-[background-color_0.2s,transform_0.2s] checked:after:content-[''] hover:cursor-pointer dark:bg-neutral-600 dark:after:bg-white dark:checked:bg-primary dark:checked:after:bg-primary disabled:cursor-default disabled:opacity-25 !focus:shadow-none"
+      className={[
+        "!bg-none mr-2 mt-[0.3rem] h-7 w-14 appearance-none rounded-[0.9rem] bg-neutral-300 before:pointer-events-none before:absolute before:h-7 before:w-5 before:rounded-full before:bg-transparent before:content-[''] after:absolute after:z-[2] after:mt-[0.2rem] after:ml-[0.2rem] after:h-5 after:w-5 after:rounded-full after:border-none after:bg-neutral-100 after:shadow-[0_0px_3px_0_rgb(0_0_0_/_7%),_0_2px_2px_0_rgb(0_0_0_/_4%)] after:transition-[background-color_0.2s,transform_0.2s] after:content-[''] checked:bg-primary checked:after:absolute checked:after:z-[2] checked:after:ml-[1.9625rem] checked:after:w-5 checked:after:rounded-full checked:after:border-none checked:after:bg-primary checked:after:shadow-[0_3px_1px_-2px_rgba(0,0,0,0.2),_0_2px_2px_0_rgba(0,0,0,0.14),_0_1px_5px_0_rgba(0,0,0,0.12)] checked:after:transition-[background-color_0.2s,transform_0.2s] checked:after:content-[''] hover:cursor-pointer dark:bg-neutral-600 dark:after:bg-white dark:checked:bg-primary dark:checked:after:bg-primary disabled:cursor-default disabled:opacity-25 !focus:shadow-none",
+        `testid:${props.testClassId}`
+      ].join(' ')}
       checked={!!props.checked}
       disabled={props.disabled}
+      data-checkbox-id={props.testClassId}
     />)
 }
--- a/tdrive/frontend/src/app/atoms/input/input-checkbox-slider.tsx
+++ b/tdrive/frontend/src/app/atoms/input/input-checkbox-slider.tsx
@@ -24,10 +24,9 @@ export const CheckboxSlider = (props: CheckboxSliderProps) => {
       // @author https://tw-elements.com/docs/react/forms/switch/
       className={[
         "!bg-none mr-2 mt-[0.3rem] h-7 w-14 appearance-none rounded-[0.9rem] bg-neutral-300 before:pointer-events-none before:absolute before:h-7 before:w-5 before:rounded-full before:bg-transparent before:content-[''] after:absolute after:z-[2] after:mt-[0.2rem] after:ml-[0.2rem] after:h-5 after:w-5 after:rounded-full after:border-none after:bg-neutral-100 after:shadow-[0_0px_3px_0_rgb(0_0_0_/_7%),_0_2px_2px_0_rgb(0_0_0_/_4%)] after:transition-[background-color_0.2s,transform_0.2s] after:content-[''] checked:bg-primary checked:after:absolute checked:after:z-[2] checked:after:ml-[1.9625rem] checked:after:w-5 checked:after:rounded-full checked:after:border-none checked:after:bg-primary checked:after:shadow-[0_3px_1px_-2px_rgba(0,0,0,0.2),_0_2px_2px_0_rgba(0,0,0,0.14),_0_1px_5px_0_rgba(0,0,0,0.12)] checked:after:transition-[background-color_0.2s,transform_0.2s] checked:after:content-[''] hover:cursor-pointer dark:bg-neutral-600 dark:after:bg-white dark:checked:bg-primary dark:checked:after:bg-primary disabled:cursor-default disabled:opacity-25 !focus:shadow-none",
-        `testid:${props.testClassId}`
+        props.testClassId ? `testid:${props.testClassId}` : ''
       ].join(' ')}
       checked={!!props.checked}
       disabled={props.disabled}
-      data-checkbox-id={props.testClassId}
     />)
 }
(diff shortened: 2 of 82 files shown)


tdrive/frontend/src/app/atoms/button/button.tsx
```diff
@@ -65,10 +66,12 @@ export const Button = (props: ButtonProps) => {
         ' inline-flex items-center px-4 py-2 border font-medium rounded-md focus:outline-none ' +
         className +
         ' ' +
-        props.className
+        props.className +
+        ' ' +
+        `testid:${props.testClassId}`
```
ericlinagora: please don't add a class for undefined, it's just clutter in the html. (this happens bellow also, but I'll only comment here).

maybe it's worth making a helper function that takes the id and returns `id ? ' testid:' + id : ''` (note the space) and call that from where it's injected into class names, this way, say we decide to change the "testid:" prefix, it would be easier, and everywhere will be protected from applying the testid:undefined class.

tdrive/frontend/src/app/atoms/input/input-checkbox-slider.tsx
```diff
@@ -15,13 +17,17 @@ export const CheckboxSlider = (props: CheckboxSliderProps) => {
...
+        `testid:${props.testClassId}`
+      ].join(' ')}
       checked={!!props.checked}
       disabled={props.disabled}
+      data-checkbox-id={props.testClassId}
```
ericlinagora: what's the purpose of this attribute ? (also only commenting once, but happens again eg: `data-input-id` )

tdrive/frontend/src/app/components/menus/menu.jsx
```diff
@@ -63,6 +63,8 @@ export default class Menu extends React.Component {
         this.props.menu,
         elementRect,
         this.props.position,
+        {},
```
ericlinagora: if you need to fill a value for an optional param, consider just using `undefined`, this instead creates a new object and could have a different meaning to that function than an undefined one

tdrive/frontend/src/app/components/menus/menus-body-layer.jsx
```diff
@@ -210,6 +210,7 @@ export default class MenusBodyLayer extends React.Component {
                           : 'skew_in_right_nobounce'
                         : 'fade_in'
                     }
+                    testClassId={item.dataMenuId}
```
ericlinagora: the mix of class, classid, id, data etc is a bit confusing.

In this example, the class id could just be menu-item, and because in this case we want both the list and to distinguish them, add the data-menu-id attribute also. Also there's no guarantee item.dataMenuId is globally unique on its own.

tdrive/frontend/src/app/views/client/body/drive/browser.tsx
```diff
@@ -232,10 +232,10 @@ export default memo(
         inPublicSharing,
       };
       return isMobile ? (
-        <DocumentRow {...commonProps} />
+        <DocumentRow {...commonProps} testClassId="browser-document-row" />
```
ericlinagora: this is great, semantically what i would have done also. But please check because using css selectors makes it easy to find by hierarchy, if there is something else in browser to select, then maybe the browser part should be on it and only document-row here. We can then select just as easily `.browser .document-row` and if there's other stuff in the browser, we can reuse that parent class

tdrive/frontend/src/app/views/client/body/drive/browser.tsx
```diff
@@ -344,8 +345,11 @@ export default memo(
                               buildFileTypeContextMenu(),
                               { x: evt.clientX, y: evt.clientY },
                               'center',
+                              {},
```
ericlinagora: default to undefined please (ok this too repeats below so I'll stop this comment on this one)

tdrive/frontend/src/app/views/client/body/drive/modals/create/index.tsx
```diff
@@ -132,6 +134,7 @@ export const CreateModal = ({
                       onClick={() =>
                         addFromUrl(app.emptyFile.url, app.emptyFile.filename || app.emptyFile.name)
                       }
+                      testClassId={app.emptyFile.name}
```
ericlinagora, line 137: I really hate dynamic values in the test class, they should only serve to pick a ui element. If many match and the difference is only data, and if we can't use the text inside (eg: for files we can, but for menu items we can't because they're translated, filenames aren't), then I think we should add separate the dynamic data into data-* attributes.

Now this specific case is annoying:
  - The default filename should probably be translated (Untilted.docx should probably be mới.docx - well, approx :) )
  - Today the actual filename is kind of used as an ID to differentiate docx from xlsx etc, so, this is correct code now

ericlinagora (changes requested):

ericlinagora:
Note - at this timestamp just closed the review because the code was updated while it was pending, so I'll restart instead of continuing on the old version

tdrive/frontend/src/app/components/file-uploads/pending-file-components/pending-file-row.tsx
```diff
@@ -191,14 +194,15 @@ export default ({ pendingFileState, pendingFile }: PropsType) => {
...
+                className="testid:pending-file-button-cancel"
               />
             </Tooltip>
           ) : (
             <div style={{ width: 32 }} />
           )}
         </Col>
       </Row>
-      <div className="file-progress-bar-container">
+      <div className="file-progress-bar-container testid:pending-file-progress-bar">
```
ericlinagora: this is one example where i think the existing classname was largely sufficient to select this

tdrive/frontend/src/app/components/menus/menus-manager.jsx
```diff
@@ -57,7 +57,7 @@ class MenusManager extends Observable {
 
     this.notify();
   }
-  async openMenu(menu, domRect, positionType, options) {
+  async openMenu(menu, domRect, positionType, options, dataMenuId) {
```
ericlinagora: if dataMenuId is the testClassId , to avoid confusion, call it the same please

tdrive/frontend/src/app/components/search-popup/parts/result-context.tsx
```diff
@@ -4,7 +4,7 @@ import { UserType } from '@features/users/types/user';
 
 export default ({ user }: { user?: UserType }) => {
   return (
-    <div className="flex overflow-hidden whitespace-nowrap text-ellipsis">
+    <div className="flex overflow-hidden whitespace-nowrap text-ellipsis testid:result-context">
```
ericlinagora: result context is super vague, please, search result context (if we interact with it in the tests at all, or container, or plural results, i'm not sure how context applies to this component name tbh)

tdrive/frontend/src/app/views/client/body/drive/browser.tsx
```diff
@@ -423,9 +433,9 @@ export default memo(
                   </Button>
                 </Menu>
                 {viewId !== 'shared_with_me' && (
-                  <Menu menu={() => onBuildContextMenu(details)}>
+                  <Menu menu={() => onBuildContextMenu(details)} testClassId="browser-menu-context">
                     {' '}
-                    <Button theme="secondary" className="ml-4 flex flex-row items-center">
+                    <Button theme="secondary" className="ml-4 flex flex-row items-center" testClassId="browser-button-context">
```
ericlinagora: say you're reading a test, it goes page.locator('.testClassId:browser-button-context'); and you didn't know you added this, it's pretty unclear right ?

tdrive/frontend/src/app/views/client/body/drive/context-menu.tsx
```diff
@@ -85,20 +85,23 @@ export const useOnBuildContextMenu = (
             !FeatureTogglesService.isActiveFeatureName(FeatureNames.COMPANY_MANAGE_ACCESS);
           const newMenuActions = [
             {
+              id: 'share',
               type: 'menu',
               icon: 'share-alt',
```
ericlinagora, line 90: if it's an id exclusively used for tests, please make it clear, but also this is one case where we can actually distinguish items the way the user would, by using the icon (name) ?

tdrive/frontend/src/app/views/client/body/drive/modals/confirm-trash/index.tsx
```diff
@@ -28,7 +28,7 @@ export const ConfirmTrashModal = () => {
   const [state, setState] = useRecoilState(ConfirmTrashModalAtom);
 
   return (
-    <Modal open={state.open} onClose={() => setState({ ...state, open: false })}>
+    <Modal className="confirm-trash-modal" open={state.open} onClose={() => setState({ ...state, open: false })}>
```
ericlinagora: `testid:`

tdrive/frontend/src/app/views/client/common/account.tsx
```diff
@@ -22,17 +22,20 @@ export default (): JSX.Element => {
       menu={[
         // user name / email
         {
+          id: 'username',
```
ericlinagora: ditto, if this is a testId, call it as such, `id` has a pretty unique meaning

tdrive/frontend/src/app/views/client/common/disk-usage.tsx
```diff
@@ -44,19 +44,19 @@ const DiskUsage = () => {
...
-                <div style={{ width: used +  '%',}} className="shadow-none flex flex-col text-center whitespace-nowrap text-white justify-center bg-green-500"></div>
+                <div style={{ width: used +  '%',}} className="shadow-none flex flex-col text-center whitespace-nowrap text-white justify-center bg-green-500 testid:disk-usage-less-than-80"></div>
               )}
               { (used >= 80  && used <= 90 )&& (
-                <div style={{ width: used +  '%',}} className="shadow-none flex flex-col text-center whitespace-nowrap text-white justify-center bg-yellow-500"></div>
+                <div style={{ width: used +  '%',}} className="shadow-none flex flex-col text-center whitespace-nowrap text-white justify-center bg-yellow-500 testid:disk-usage-80-to-90"></div>
               )}
 
-              <div style={{ width: (100 - used) +  '%' }} className="shadow-none flex flex-col text-center whitespace-nowrap text-white justify-center bg-grey-500"></div>
+              <div style={{ width: (100 - used) +  '%' }} className="shadow-none flex flex-col text-center whitespace-nowrap text-white justify-center bg-grey-500 testid:disk-usage-100"></div>
```
ericlinagora: i think technically this one is showing the free space

ericlinagora (changes requested):
overall this is a ton of work well done, my issue is that it was too early/quickly done, and i see testids more as a last resort, only better than using labels/roles or data-test-id. If there's any other reliable way to identify something we should probably do that. I think after the mini fixes we'll take it in, then do a second pass review with QA with actual script test cases etc.

tdrive/frontend/src/app/components/menus/menus-manager.jsx
ericlinagora: (but if it should become `data-menu-id=""` then great)

tdrive/frontend/src/app/atoms/input/input-checkbox-slider.tsx
lethemanh: It has been removed

tdrive/frontend/src/app/atoms/button/button.tsx
lethemanh: It's been resolved

tdrive/frontend/src/app/views/client/body/drive/browser.tsx
lethemanh: I understand your idea, but this name has a rule behind, it is combination of component name and element name. Also, it helps to reduce the number of parent classes needed for selector

tdrive/frontend/src/app/views/client/body/drive/modals/create/index.tsx
lethemanh: Actually, QA want each app in this modal has an unique test id so I have to use its name, but also I checked that `app.emptyFile.name` is key for translation so I think it can be use as test id in this case.
CleanShot 2024-12-11 at 09 55 46@2x

tdrive/frontend/src/app/components/file-uploads/pending-file-components/pending-file-row.tsx
lethemanh: As I mentioned in this [comment](https://github.com/linagora/twake-drive/pull/764#discussion_r1879243652)

tdrive/frontend/src/app/components/menus/menus-manager.jsx
lethemanh: It's been renamed

tdrive/frontend/src/app/views/client/common/account.tsx
lethemanh: It's been renamed

tdrive/frontend/src/app/views/client/common/disk-usage.tsx
lethemanh: It's been renamed

tdrive/frontend/src/app/views/client/body/drive/modals/confirm-trash/index.tsx
lethemanh: Updated

tdrive/frontend/src/app/views/client/body/drive/context-menu.tsx
lethemanh: I've changed into `testClassId`

tdrive/frontend/src/app/views/client/body/drive/browser.tsx
lethemanh: It's been renamed

tdrive/frontend/src/app/components/search-popup/parts/result-context.tsx
lethemanh: It's been renamed

tdrive/frontend/src/app/views/client/body/drive/modals/create/index.tsx
ericlinagora: > Today the actual filename is kind of used as an ID to differentiate docx from xlsx etc, so, this is correct code now

Like i wrote, this one is annoying, and we'll probably have to change it in the future, but now, this is as good as possible

tdrive/frontend/src/app/views/client/body/drive/browser.tsx
ericlinagora: But we don't want to reduce the number of parent selectors, and this means:

1) we have to repeat multiple times the parent's name
2) it's not composable if this is reused in multiple places
3) it the parent is repeated then using the hierarchy ensures we get the consistent set of children
4) if we don't need the parent in the selector because this isn't reused, repeated, etc ; then there's still no reason to repeat the parent selector on each child

please don't mix parent and child element, the DOM will let us choose how much hierarchy we want to specify and change it easily without changing the code if we need to

tdrive/frontend/src/app/components/file-uploads/pending-file-components/pending-file-row.tsx
ericlinagora: yes but we should see the test id as a last resort, when we can't use selectors that would already exist. This is mostly for anything that doesn't have an id and has only tailwind style classes. But in this case, it was just as easy to write the test without adding this. That's why I wanted to add the test ids only at the request of QA by necessity.

tdrive/frontend/src/app/views/client/body/drive/browser.tsx
ericlinagora: Here is one example to clarify why this is a very bad idea, beyond just having to repeat code, being un-composable, breaching concern separation etc. Imagine we want to read each person in the list:

```html
<li id="person">
  <div id="name">name 1</div>
  <div id="phone">123</div>
</li>
<li id="person">
  <div id="name">name 2</div>
  <div id="phone">123</div>
</li>
```

versus

```html
<li>
  <div id="person-name">name 1</div>
  <div id="person-phone">123</div>
</li>
<li>
  <div id="person-name">name 2</div>
  <div id="person-phone">123</div>
</li>
```

see how much more complicated it would be to figure out which phone number goes with which name ? youd end up walking up and back down the tree and be specific to the html structure.

now imagine we want to reuse the name+phone for companies in another list.

now imagine we want to rename "person" into "employee" you have to change it at multiple places instead of one.

ericlinagora: because of the size of this PR, and the rapidly increasing risk of conflict, let's merge this for now

ericlinagora (approved):
Approval to figure this out without the pending huge PR merge since it's benign

PR #771, opened 2024-12-16
🐛 Fix ORM query builder with $in option for postgres

### LINKED ISSUE #770: [Bug] User search in the access modal is not working with Postgres
**Describe the bug**
On the "manage access" trying to search for users is not working.

**To Reproduce**
1. Go to https://drive.stg.lin-saas.com/
2. Toggle the context menu for any item and chosse "manage access"
3. Try to search for an existing user

MontaGhanmy:


ericlinagora (changes requested):
There's no valid reason to make a query with an empty `$in: []`. Ignoring this condition all together can lead to data leaks. At best the expected result would be 0 rows since nothing is IN that list. But throwing an error and fixing it at the call level, figuring out what lead to this being empty is more important.

ericlinagora (approved):

PR #775, opened 2024-12-26
✨ Implement API update user

### LINKED ISSUE #773: Create a new API "update' in user controller
This is the request params, and information we can update
```
export interface UserObject {
  email: string;
  picture: string;
  first_name: string;
  last_name: string;
  full_name: string;
}
```

if the email is set and it's different from the current one, we should update login information with `ExternalUser`(entity that we use in case of OIDC user provider)

in e2e test we should test that user is able to login after email change

lethemanh:


tdrive/backend/node/src/services/user/services/users/service.ts
```diff
@@ -92,8 +111,37 @@ export class UserServiceImpl {
...
+    instance.email_canonical = user.email || instance.email_canonical;
+    instance.first_name = user.first_name || instance.first_name;
+    instance.last_name = user.last_name || instance.last_name;
+    instance.picture = user.picture || instance.picture;
+    instance.creation_date = !isNumber(instance.creation_date)
+      ? Date.now()
+      : instance.creation_date;
+    if (isChangeEmail) {
+      instance.username_canonical = (
+        user.email.replace(/[^a-z0-9_-]/, "") || ""
```
ericlinagora: I think this is copied from https://github.com/linagora/twake-drive/blob/df5c472d7b5fdcb15db23b5f8b16f327c9c7f861/tdrive/backend/node/src/services/user/services/users/service.ts#L85 , couldn't we have this logic in a single place ? in case the regex changes for example ?

MontaGhanmy (approved):

PR #777, opened 2025-01-08
🐛 Fix bug infinite switching preview files when using shortcut keys

lethemanh:


tdrive/frontend/src/app/views/client/viewer/drive-preview.tsx
```diff
@@ -25,6 +25,8 @@ import Controls from './controls';
 interface DrivePreviewProps {
   items: DriveItem[];
 }
+let currentItemIndex: any;
```
ericlinagora: can't we type this as number | undefined ? i like to avoid any as much as possible, it kind defeats the whole point of TS in the first place

tdrive/frontend/src/app/views/client/viewer/drive-preview.tsx
```diff
@@ -67,16 +81,14 @@ export const DrivePreview: React.FC<DrivePreviewProps> = ({ items }) => {
...
+    addShortcut({ shortcut: 'Left', handler: handleSwitchLeft });
 
     return () => {
-      removeShortcut({ shortcut: 'Right', handler: handleSwitchRight });
-      removeShortcut({ shortcut: 'Left', handler: handleSwitchLeft });
+      removeShortcut({ shortcut: 'Right' });
+      removeShortcut({ shortcut: 'Left' });
     };
-  }, [status]);
+  }, [JSON.stringify([...items])]);
```
ericlinagora: This seems risky, i dont think there's a garantee about things like key orders in the json output, and would build quite a big string, do we need the deep equality here ? could testing on items.map(({id}) => id) or similar do the same job ?

tdrive/frontend/src/app/views/client/viewer/drive-preview.tsx
```diff
@@ -67,16 +81,14 @@ export const DrivePreview: React.FC<DrivePreviewProps> = ({ items }) => {
...
-      addShortcut({ shortcut: 'Left', handler: handleSwitchLeft });
-    }
+    addShortcut({ shortcut: 'Right', handler: handleSwitchRight });
+    addShortcut({ shortcut: 'Left', handler: handleSwitchLeft });
 
     return () => {
-      removeShortcut({ shortcut: 'Right', handler: handleSwitchRight });
-      removeShortcut({ shortcut: 'Left', handler: handleSwitchLeft });
+      removeShortcut({ shortcut: 'Right' });
+      removeShortcut({ shortcut: 'Left' });
```
ericlinagora, line 89: Because they are global, doesn't this run the risk of interfering with another component also using the left and right shortcuts ? I understand handleSwitchLeft has a different value during a second run, can't put them in a useState or similar to keep the identical instances ?

lethemanh: yes, I've tried to put `handleSwitchLeft` and `handleSwitchRight` into the state to keep the instances but it did not work

ericlinagora (approved):

MontaGhanmy (approved):

PR #778, opened 2025-01-08
🐛 Fixing pagination for "Shared with me"/ES

### LINKED ISSUE #774: Shared with me in STG SAAS is empty
After sharing a file with another user, the user does get the e-mail, which does link to the "shared with me" section, but the file is not there/shared with me is empty. Could be deployment related and not strictly a bug

MontaGhanmy:


tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -133,6 +133,24 @@ export class DocumentsService {
...
+        };
+      } else {
+        // Clear pagination if no nextPage token is provided
+        options.pagination = {
+          limitStr: options.pagination?.limitStr,
+        };
+      }
+    }
+
+    console.log("🚀🚀 loading with:: ", options);
```
ericlinagora: if we need the log, and we go through the logger with a structured output ?

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -133,6 +133,24 @@ export class DocumentsService {
       options.sort = this.getSortFieldMapping(options.sort);
     }
 
+    // Handle pagination differently for non-MongoDB platforms
+    if (globalResolver.platformServices.search.type !== "mongodb") {
```
ericlinagora: i'd prefer to avoid the magic string in the generic part of the code, we could do it by just moving this logic into the connector, and have an empty implem for mongo. At a minimum, if it's not the other expected string, then throw a notimplementederror

tdrive/frontend/src/app/features/drive/hooks/use-drive-actions.tsx
```diff
@@ -70,11 +70,12 @@ export const useDriveActions = (inPublicSharing?: boolean) => {
           let pagination = await snapshot.getPromise(DriveItemPagination);
 
           if (resetPagination) {
-            pagination = { page: 0, limit: pagination.limit };
+            pagination = { page: 0, limit: pagination.limit, nextPage: pagination.nextPage };
             set(DriveItemPagination, pagination);
           }
+          let details: any;
```
ericlinagora: instead of any can something like `Awaited<ReturnType<typeof DriveApiClient["browse"]>> | undefined` ?

tdrive/backend/node/src/core/platform/services/search/index.ts
```diff
@@ -96,4 +97,30 @@ export default class Search extends TdriveService<SearchServiceAPI> {
   api(): SearchServiceAPI {
     return this;
   }
+
+  handlePagination(options: SearchDocumentsOptions): void {
+    if (this.type === "mongodb") {
```
ericlinagora, line 102: this a lot better, but still, if we just move this fuction to ESAndOpenSearch and MongosearchService there's no need to look at the magic strings, it's future proof, etc

ericlinagora: the main idea of this whole class is only have mutual code and leave the specialisations to the adapters

MontaGhanmy: It doesn't have to be moved that deep into the connector and we're looking at the magic string either ways when initializing the service.

PR #779, opened 2025-01-15
Add testid into checkable icon, file version rows

lethemanh:


tdrive/frontend/src/app/views/client/body/drive/modals/create/index.tsx
```diff
@@ -135,7 +135,7 @@ export const CreateModal = ({
                       onClick={() =>
                         addFromUrl(app.emptyFile.url, app.emptyFile.filename || app.emptyFile.name)
                       }
-                      testClassId={app.emptyFile.name}
+                      testClassId={app.emptyFile.name.replaceAll(' ' ,'-').toLocaleLowerCase()}
```
ericlinagora, line 138: no dynamic testid's please, the filename can be read from the correspondin span child

tdrive/frontend/src/app/views/client/body/drive/modals/versions/index.tsx
```diff
@@ -99,17 +99,18 @@ const VersionModalContent = ({ id }: { id: string }) => {
...
               'flex flex-row items-center border -mt-px px-4 py-3 cursor-pointer hover:bg-zinc-500 hover:bg-opacity-10 ' +
               (index === 0 ? 'rounded-t-md ' : '') +
-              (index === (versions || []).length - 1 ? 'rounded-b-md ' : '')
+              (index === (versions || []).length - 1 ? 'rounded-b-md ' : '') +
+              'testid:version-row'
             }
           >
             <div className="grow text-ellipsis whitespace-nowrap overflow-hidden">
-              <Base>{version.file_metadata.name}</Base>
+              <Base className="testid:version-name">{version.file_metadata.name}</Base>
```
ericlinagora: there's no need to repeat the parent name, finding this row would be easy with `.version-row .name` too

tdrive/frontend/src/app/atoms/input/input-select.tsx
```diff
@@ -33,6 +33,7 @@ export function Select(props: InputProps) {
       style={styles}
       className={inputClassName + ' ' + props.className}
       {..._.omit(props, 'label', 'className', 'size')}
+      data-value={props.value}
```
ericlinagora: try and read this is as a user would, we should have as little as possible that is only needed by tests

ericlinagora (changes requested):

tdrive/frontend/src/app/views/client/body/drive/modals/create/index.tsx
ericlinagora: this is the only office template file names

ericlinagora (approved):

PR #780, opened 2025-01-21
✨ Big folder upload

MontaGhanmy:
output


tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -371,6 +371,7 @@ export class DocumentsService {
     content: Partial<DriveFile>,
     version: Partial<FileVersion>,
     context: DriveExecutionContext,
+    tmp = false,
```
ericlinagora: Seems we can just use content.is_in_trash here instead of the added param, makes it more obvious the impact on the model than `tmp`

shepilov (approved):

PR #781, opened 2025-01-22
🐛 Check permission for API update user

### LINKED ISSUE #773: Create a new API "update' in user controller

lethemanh:


ericlinagora (changes requested):
Can we check this company admin persmission stuff in an drive backend e2e test please ?

ericlinagora (approved):
Expediency

PR #788, opened 2025-01-23
🐛 [in brackets] back: immediately start streaming downloading of zip #783

### LINKED ISSUE #783: Bug ERR_CONNECTION_CLOSE when downloading a zip of a folder with a great many number of files
Note: update ticker 3954 in internal system too.

ericlinagora:
When downloading a folder with a great many entries below, time to first byte can cause a tcp disconnection for timeout by a reverse proxy, this starts the download asap, and ignores unreadable entries because of AV or rights. It tries to generate a valid zip even if some entries are missing for those reasons

### DIFF
--- a/tdrive/backend/node/src/services/documents/utils.ts
+++ b/tdrive/backend/node/src/services/documents/utils.ts
@@ -345,6 +345,9 @@ export const addDriveItemToArchive = async (
   id: string,
   entity: DriveFile | null,
   archive: archiver.Archiver,
+  beginArchiveTransmit:
+    | "ONLY_SET_THIS_VALUE_IF_ALREADY_CALLED"
+    | ((archive: archiver.Archiver) => void),
   repository: Repository<DriveFile>,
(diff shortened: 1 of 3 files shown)


tdrive/backend/node/src/services/documents/utils.ts
```diff
@@ -345,6 +345,9 @@ export const addDriveItemToArchive = async (
   id: string,
   entity: DriveFile | null,
   archive: archiver.Archiver,
+  beginArchiveTransmit:
```
shepilov, line 348: I love this, string or function, do you miss void*?)

shepilov (approved):

tdrive/backend/node/src/services/documents/utils.ts
ericlinagora: lol, no that's just object. Here i want to make sure it's not called with a falsy/null/undef value if it isn't done only for the reason of passing the call having happened down the recursion

PR #796, opened 2025-02-11
♻️ Runnable performance tests

shepilov:


tdrive/backend/node/test/performance/files/example.js
```diff
@@ -0,0 +1,15 @@
...
+import { sleep } from 'k6';
+
+export const options = {
+  iterations: 10,
+};
+
+// The default exported function is gonna be picked up by k6 as the entry point for the test script. It will be executed repeatedly in "iterations" for the whole duration of the test.
+export default function () {
+  // Make a GET request to the target URL
+  http.get('https://test-api.k6.io');
```
ericlinagora, line 11: ?

ericlinagora (approved):

tdrive/backend/node/test/performance/files/example.js
shepilov: Minimal "hello world" if somebody will try to run tests locally and it won't work

PR #820, opened 2025-02-24
799 user deletion

### LINKED ISSUE #799: User deletion
Delete all data related to a user from a secured internal end-point.

- [] Prevent login by deleted user
- [] Search and access rights cannot find deleted user
- [] Remove data from S3 progressively
- [] Remove data from search
- [] Remove data from DB

[Internal documentation](https://www.notion.so/linagora/Delete-TWP-account-19d62718bad180d39942ce3b1a419d40)

ericlinagora:
## Devops concerns

- set `ADMIN_ENDPOINT_SECRET` env to a secret string
- ensure `/admin/*` urls aren't accessible from the public
- Storage configuration:
  - Single S3: set  env var to `STORAGE_S3_OVERRIDE_DISABLE_REMOVE_FOR_USER_ACCOUNT_DELETION=true`, any other value than the string `true` is false (and no override is set, so deletion from S3 is blocked in all cases)
  - OneOf S3: set the json key `overrideDisableRemoveForUserAccountDeletion` to `true` of each S3

## Caller concerns

- POST `/admin/api/user/delete` with a body `{ secret, userId, deleteData }` where `deleteData` is `true` to begin/advance actual data deletion. Result is a json body with `{ status: "failed" | "deleting" | "done"}`. Caller should expect this endpoint to timeout, and ignore it. This should be called on a periodic basis until all users have returned `{status: "done"}`. If the result is `failed` or not 200 trigger a devops alarm for manual review. For dev/test purposes, using `userId: "e2e_simulate_timeout"` will not respond causing a timeout on the caller, so the caller may specifically ignore that error as it is expected because of the potential volume of operations for the deletion to complete.
- POST `/admin/api/user/pending` with a body `{ secret }`, response will be a list of `userId`s and their deletion start date in the format `[ [ "userid_1", 123 ], [ "userid_2", 321 ] ]` that are marked for deletion but not deleted yet.

## Call flow
- At user deletion requestion:
  - POST `/admin/api/user/delete` to delete a user, it will either return
    - `{ status: "done" }` : This is the end, the user was succesfully deleted completely in the time of one call
    - `{ status: "failed" }`: Alert sysadmin, deleting failed, might have even failed to add to pending users for deletion. Store in caller for re-try.
    - `{ status: "deleting" }` or a request timeout *(test with `userId: "e2e_simulate_timeout"` to get exact caller error in that case)* : User deletion is scheduled and incomplete, use the following flow
- Periodic polling:
  - POST `/admin/api/user/pending`: For each returned used, if age (unix epoch) is too old, flag an alert and skip. Otherwise call POST `/admin/api/user/delete` for that user again until succesful (`{status: "done" }`).

### DIFF
--- a/tdrive/backend/node/config/custom-environment-variables.json
+++ b/tdrive/backend/node/config/custom-environment-variables.json
@@ -102,7 +102,8 @@
       "useSSL": "STORAGE_S3_USE_SSL",
       "accessKey": "STORAGE_S3_ACCESS_KEY",
       "secretKey": "STORAGE_S3_SECRET_KEY",
-      "disableRemove": "STORAGE_S3_DISABLE_REMOVE"
+      "disableRemove": "STORAGE_S3_DISABLE_REMOVE",
+      "overrideDisableRemoveForUserAccountDeletion": "STORAGE_S3_OVERRIDE_DISABLE_REMOVE_FOR_USER_ACCOUNT_DELETION"
     }
   },
   "email-pusher": {
--- a/tdrive/backend/node/config/custom-environment-variables.json
+++ b/tdrive/backend/node/config/custom-environment-variables.json
@@ -16,6 +16,9 @@
   "logger": {
     "level": "LOG_LEVEL"
   },
+  "admin": {
+    "endpointSecret": "ADMIN_ENDPOINT_SECRET"
+  },
   "diagnostics": {
     "skipKeys": {
       "__name": "DIAG_SKIP_KEYS",
--- a/tdrive/backend/node/src/core/platform/services/admin/controller/delete-user-controller.ts
+++ b/tdrive/backend/node/src/core/platform/services/admin/controller/delete-user-controller.ts
@@ -1,58 +1,37 @@
 import gr from "../../../../../services/global-resolver";
-import { getLogger } from "../../../../../core/platform/framework";
-import User, { TYPE as UserType } from "../../../../../services/user/entities/user";
-import type { DatabaseServiceAPI } from "../../database/api";
+
 import type { ExecutionContext } from "../../../../platform/framework/api/crud-service";
-import type { SearchServiceAPI } from "../../search/api";
-import {
-  DriveFile,
-  TYPE as DriveFileType,
-} from "../../../../../services/documents/entities/drive-file";
-import { File } from "../../../../../services/files/entities/file";
-import {
-  FileVersion,
-  TYPE as FileVersionType,
-} from "../../../../../services/documents/entities/file-version";
-import ExternalUser, {
-  TYPE as ExternalUserType,
-} from "../../../../../services/user/entities/external_user";
-import CompanyUser, {
-  TYPE as CompanyUserType,
-} from "../../../../../services/user/entities/company_user";
-const FileType = "files";
 
-const logger = getLogger("AdminDeleteUserController");
+import { adminLogger, buildUserDeletionRepositories } from "../utils";
 
-/**
- * Create all repositories required for deleting a user
- * @deprecated Do not use this outside of this file, it is exported exclusively for e2e tests
- */
-export async function buildUserDeletionRepositories(
-  db: DatabaseServiceAPI,
-  search: SearchServiceAPI,
-) {
-  return {
-    driveFile: await db.getRepository<DriveFile>(DriveFileType, DriveFile),
-    file: await db.getRepository<File>(FileType, File),
-    fileVersion: await db.getRepository<FileVersion>(FileVersionType, FileVersion),
+// /**
+//  * Create all repositories required for deleting a user
+//  * @deprecated Do not use this outside of this file, it is exported exclusively for e2e tests
+//  */
+// export async function buildUserDeletionRepositories(
+//   db: DatabaseServiceAPI,
+//   search: SearchServiceAPI,
+// ) {
+//   return {
+//     driveFile: await db.getRepository<DriveFile>(DriveFileType, DriveFile),
+//     file: await db.getRepository<File>(FileType, File),
+//     fileVersion: await db.getRepository<FileVersion>(FileVersionType, FileVersion),
 
-    user: await db.getRepository<User>(UserType, User),
-    companyUser: await db.getRepository<CompanyUser>(CompanyUserType, CompanyUser),
-    externalUser: await db.getRepository<ExternalUser>(ExternalUserType, ExternalUser),
+//     user: await db.getRepository<User>(UserType, User),
+//     companyUser: await db.getRepository<CompanyUser>(CompanyUserType, CompanyUser),
+//     externalUser: await db.getRepository<ExternalUser>(ExternalUserType, ExternalUser),
 
-    // group_entity
-    // group_user
-    // missed_drive_files
-    // session
-    // user
-    // user_online
+//     // company: await db.getRepository<Company>(CompanyType, Company),
+//     // missed_drive_files
+//     // session
+//     // user_online
 
-    search: {
-      driveFile: await search.getRepository<DriveFile>(DriveFileType, DriveFile),
-      user: await search.getRepository<User>(UserType, User),
-    },
-  };
-}
+//     search: {
+//       driveFile: await search.getRepository<DriveFile>(DriveFileType, DriveFile),
+//       user: await search.getRepository<User>(UserType, User),
+//     },
+//   };
+// }
 
 export class AdminDeleteUserController {
   private constructor(
@@ -77,7 +56,8 @@ export class AdminDeleteUserController {
         if (existingUser.delete_process_started_epoch > 0) return "deleting";
       }
     } catch (err) {
-      logger.error({ err, userId }, "User deletion error");
+      adminLogger.error({ err, userId }, "User deletion error");
+      console.log(err); //TODO: NONONO
       return "failed";
     }
     return "done";
--- a/tdrive/backend/node/src/core/platform/services/admin/controller/delete-user-controller.ts
+++ b/tdrive/backend/node/src/core/platform/services/admin/controller/delete-user-controller.ts
@@ -4,45 +4,13 @@ import type { ExecutionContext } from "../../../../platform/framework/api/crud-s
 
 import { adminLogger, buildUserDeletionRepositories } from "../utils";
 
-// /**
-//  * Create all repositories required for deleting a user
-//  * @deprecated Do not use this outside of this file, it is exported exclusively for e2e tests
-//  */
-// export async function buildUserDeletionRepositories(
-//   db: DatabaseServiceAPI,
-//   search: SearchServiceAPI,
-// ) {
-//   return {
-//     driveFile: await db.getRepository<DriveFile>(DriveFileType, DriveFile),
-//     file: await db.getRepository<File>(FileType, File),
-//     fileVersion: await db.getRepository<FileVersion>(FileVersionType, FileVersion),
-
-//     user: await db.getRepository<User>(UserType, User),
-//     companyUser: await db.getRepository<CompanyUser>(CompanyUserType, CompanyUser),
-//     externalUser: await db.getRepository<ExternalUser>(ExternalUserType, ExternalUser),
-
-//     // company: await db.getRepository<Company>(CompanyType, Company),
-//     // missed_drive_files
-//     // session
-//     // user_online
-
-//     search: {
-//       driveFile: await search.getRepository<DriveFile>(DriveFileType, DriveFile),
-//       user: await search.getRepository<User>(UserType, User),
-//     },
-//   };
-// }
-
 export class AdminDeleteUserController {
-  private constructor(
-    private readonly repos: Awaited<ReturnType<typeof buildUserDeletionRepositories>>,
-  ) {}
-  public static async create() {
-    return new AdminDeleteUserController(
-      await buildUserDeletionRepositories(gr.database, gr.platformServices.search),
-    );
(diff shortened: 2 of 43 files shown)


tdrive/backend/node/src/core/platform/services/admin/controller/delete-user-controller.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import type { ExecutionContext } from "../../../../platform/framework/api/crud-service";
+
+import { adminLogger, buildUserDeletionRepositories } from "../utils";
+
+export class AdminDeleteUserController {
+  private _repos: Awaited<ReturnType<typeof buildUserDeletionRepositories>>;
+  private async getRepos() {
+    if (!this._repos)
+      this._repos = await buildUserDeletionRepositories(gr.database, gr.platformServices.search);
```
shepilov, line 11: Why do you need this strange way of accessing repos?
And also using repos inside controller? UserService?

tdrive/backend/node/src/core/platform/services/admin/service.ts
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  /**
+   * Delete all the provided storage path data in batches.
+   * Returns `true` if all (or empty) operations returned `true`.
+   * Ignores (logs) errors, but cause a return of `false`.
+   *
+   * @warn No business rules applied, this is unsafe with sync, use only for final deletion.
+   */
+  private async deleteS3Paths(
```
shepilov, line 35: You can delete all files in S3 at once, as far as we have hierarchy there `../<COMPANY_ID>/<USER_ID>/...`, the same for local storage

tdrive/backend/node/config/custom-environment-variables.json
```diff
@@ -16,6 +16,9 @@
   "logger": {
     "level": "LOG_LEVEL"
   },
+  "admin": {
```
shepilov, line 19: Could you also please add a list of errors, that should be added to the monitoring? Right now, all alerts that we have right now, are bound to the error type

tdrive/backend/node/src/services/console/clients/remote.ts
```diff
@@ -328,6 +328,16 @@ export class ConsoleRemoteClient implements ConsoleServiceClient {
     }
   }
 
+  async userWasDeletedForceLogout(userId: string) {
+    const sessionRepository = gr.services.console.getSessionRepo();
+    if (!sessionRepository) return;
+    const sessions = (await sessionRepository.find({ sub: userId })).getEntities();
+    for (const session of sessions) {
```
shepilov: I'm not sure we are deleting anything from sessions, and at this point you are updating all the session, also those which were already revoked

tdrive/backend/node/src/core/platform/services/admin/utils.ts
```diff
@@ -0,0 +1,158 @@
...
+ * @param {(item: DriveFile, children: undefined | T[], parents: DriveFile[]) => Promise<T>} map - The mapping function to process each drive item.
+ *   - `item`: The current drive item being processed.
+ *   - `children`: The results of processing the children of the current drive item, or `undefined` if the item is not a directory.
+ *   - `parents`: An array of parent drive items leading to the current item.
+ * @returns {Promise<T[]>} A promise that resolves to an array of results produced by the `map` function.
+ */
+export async function descendDriveItemsDepthFirstRandomOrder<T>(
+  repos: TUserDeletionRepos,
+  parent_id: string,
+  map: (item: DriveFile, children: undefined | T[], parents: readonly DriveFile[]) => Promise<T>,
```
shepilov, line 143: Very meaningful name "map", it helps to understand what it should do, the same with docs for it

tdrive/backend/node/src/core/platform/services/admin/utils.ts
```diff
@@ -0,0 +1,158 @@
...
+) {
+  async function descend(parent_id: string, parents: readonly DriveFile[]): Promise<T[]> {
+    const items = (await repos.driveFile.find({ parent_id })).getEntities();
+    fisherYattesShuffleInPlace(items);
```
shepilov, line 147: Just why?

tdrive/backend/node/config/custom-environment-variables.json
ericlinagora: There aren't really errors at this point that should cause alerts. That should be in the caller if the user failed for long enough to delete, until then errors are tolerable and the operation should keep beeing retried. Nearly all logging relating to the deletion is in its own logger child, so when we do decide to look into a given user manually, it should be easy enough

tdrive/backend/node/src/core/platform/services/admin/controller/delete-user-controller.ts
ericlinagora: If I don't lazy load it, some of the entries aren't initialised yet or something with platform, and I get undefined if I do it at init.

Just using it for trivial raw finds, seems more readable to me than a couple of extra functions with just a find call in them, it's not like this class was too cluttered. If you want yeah I extract the one lines to one line functions, the rest ("deleting" vs "done" etc) is web api stuff ?

tdrive/backend/node/src/core/platform/services/admin/service.ts
ericlinagora: I have not found a way in S3 api to delete all objects with a common prefix, but yeah if you know of one it would be better than all this

tdrive/backend/node/src/core/platform/services/admin/utils.ts
ericlinagora: It's basically the same as the callback to Array.map, how would you call and describe it ?

tdrive/backend/node/src/core/platform/services/admin/utils.ts
ericlinagora: It's a probabilistic approach, same reason as in OO connector: 0) in case the process gets stuck on a given item for some reason related to it, this means next retry will still do progress. 1) in case the poller or other call the delete multiple concurrent times, this increases the chances they won't just mirror each other and should still individually progress as possible given they'll go down different branches of the hierarchy

tdrive/backend/node/src/services/console/clients/remote.ts
ericlinagora: couldn't filter on revoked_at: 0, so updated to only save the necessary ones. We should probably have some cleanup method for this table though

tdrive/backend/node/src/core/platform/services/admin/controller/delete-user-controller.ts
shepilov: yes, that's why it's better to put in service and have a proper platform lifecycle

tdrive/backend/node/src/core/platform/services/admin/service.ts
shepilov: yes, it look like there is no way, only list with prefix and then manually delete

tdrive/backend/node/config/custom-environment-variables.json
shepilov: But if you know that exception was thrown during operation, there is no need to rely on timeout on the client side

tdrive/backend/node/src/core/platform/services/admin/controller/delete-user-controller.ts
ericlinagora: It was in the service lifecycle before, i changed it to this because at service init, some of the repos weren't resolving

tdrive/backend/node/config/custom-environment-variables.json
ericlinagora: I don't know there will be an exception, the timeout is because I expect this whole process to easily last longer than an http request even without errors. Errors during the run are handled this way:
- Some are not detectable, so just try and move on (eg. delete on s3 always succeeds even if the file doesn't exist)
- Some are partial blockers for some things (eg, if there's an issue deleting a file, we can't delete all its parent folders because we need them to find the file next time).
- Some are not blockers, for ex. removing from the search index, if it fails, it's survivable because the search results are joined with db results, so they won't appear anyway and aren't worth getting stuck on retrying
- Some are fatal, in that case the endpoint returns the error status (not timeout)

All caught errors are logged, there just aren't really any that should cause an alert until many retries failed, which is the caller's reponsibility

tdrive/backend/node/src/core/platform/services/admin/utils.ts
shepilov: 0) We are not interested in partial progress, it will give us nothing
1) You are doing it in the controller by checking delete_process_started_epoch
So I don't think this code is relevant

ericlinagora: 0) it's always good to have and costs nothing, and i think it's still worth it so a human intervention later has least to do
1) not sure what you mean, if you call the delete user multiple times when it wasn't finished deleting, it will keep trying, the delete_process_started_epoch is not atomic, it's just indicative to detect ongoing and too long running deletions

tdrive/backend/node/src/core/platform/services/admin/utils.ts
shepilov: That's why I'm asking, it's not actually a map, inside this "map" you are actually performing a delete action.
And in case of user deletion, its a "delete" function. But you are using it also to dump items in cmd function. And in this case it will be "dump". 
So for me, when you name something map, you are interested in the result of this action function, and not in the action that will be done to transform. 

Also, type of children - T, and actually you don't need this generic, it's boolean or undefined

tdrive/backend/node/src/core/platform/services/admin/service.ts
```diff
@@ -0,0 +1,274 @@
...
+    logger: pino.Logger,
```
shepilov, line 36: why not to crate logger with the same name in service?

tdrive/backend/node/src/core/platform/services/admin/utils.ts
shepilov: 0. Why least? human actions are the same no matter how many files you deleted before
1. I mean this code 
```
if (existingUser?.deleted) {
        if (existingUser.delete_process_started_epoch > 0) return "deleting";
}
```
it's already taken some time to understand why you randomize the array before deleting, and if the code prevents running two requests consecutively, then remove this array shaker please

tdrive/backend/node/src/core/platform/services/admin/service.ts
ericlinagora: because the logger instance is specific to one user deletion operation, not the service's lifetime

tdrive/backend/node/src/core/platform/services/admin/utils.ts
ericlinagora: it's a mapping function for resolving child items into an array of something, so that when the visitor pattern leaves the accepted node, parent node can behave depending on child leaves, and itself be later mapped into the children of it's own parent when that iteration is finished. the `descendDriveItemsDepthFirstRandomOrder` knows nothing about deleting or dumping, it's just descends depth first, collecting results as it exhausts depths. That's why the result is a generic, this function doesn't know nor care about that type, only the map function will care about it because it gets its children. Maybe a better name would be visitor pattern, with enter accept and leave, but i had no reason to specially distinguish leaves from nodes here so just made it simpler.

tdrive/backend/node/src/core/platform/services/admin/utils.ts
ericlinagora: 0) If the deletion process fails for too long, our only way out is to create an alert, and have a human go look at what failed and why, probably do a bit of manual cleanup then for compliance. At that point, the less stuff left the better. Before that point, this garantees progress at each run, say there's an error deleting the first item and it stops there, next retry will do the same, so the retries become exclusively useful for timeouts, none of the other errors. This ensures some degree of resilience. It's not perfect, but it is sure more reliable than not doing it, and again, costs one array sort and a 7 line function.
1) That `if` is simply to return to the client in the case the process didn't finish succesfully. If you look above it, it waits for `anonymizeAndDelete` which retries the deletion. Again this field is not meant for concurrent stuff at all, it's not atomic in any way, not even approximately (i'm not talking details of ACID, I'm saying that's not its purpose at all and if we want to implement locks it would be an additional separate field).

shepilov: 0. And the usefulness of this function costs much less than the time we spend together discussing it, and my surprise when I saw it
1. Yes, you are right, call of the "anonymizeAndDelete" is above. And I think we should move this if above if we suppose that a client will call the same function to check status of the deletion request. Or to have another endpoint to check a status of the request. if the array of root items are the same in both calls(what is very probable in DFS) you will try to call "remove" function anyway for each item in the root items array. 

And it's also one more point, when you have 7 rows function call inside if parentesis it’s not easy to read.

tdrive/backend/node/src/core/platform/services/admin/service.ts
shepilov: Why do you need it? we have request_id and user_id in the context

ericlinagora: we don't have user_id in the context it's not the user making the call, this adds it. Also this allows filtering on all the deletion efforts for a given user exclusively

shepilov (approved):

PR #838, opened 2025-03-03
🐛 Fix bug unable to interact with upload viewer when uploading new file's version

### LINKED ISSUE #823: [Bug][new Loading popup] Could not interact with uploading popup during adding a new version
**To Reproduce**
1. Go to My drive
2. Click on ... of an uploaded file
3. Choose version
4. Add new version
5. Try to interact with uploading popup during adding a new version. Example: try to cancel or pause the upload
6. See that could not interact with uploading popup during adding a new version 

**Expected behavior**A clear and concise description of what you expected to happen.
Could interact with uploading popup during adding a new version 

**Screenshots**
https://github.com/user-attachments/assets/2148e5ae-5f7f-4c7e-b383-a7711c9b695b

**Desktop (please complete the following information):**
 - QA, Linsaasstag

lethemanh:


tdrive/frontend/src/app/components/portal/index.tsx
```diff
@@ -0,0 +1,22 @@
+import { ReactNode, useEffect } from "react";
+import ReactDOM from "react-dom";
+
+const Portal = ({ children, className }: { children: ReactNode; className?: string }) => {
```
shepilov, line 4: why do we need such a tricky method to create div?

lethemanh: The `UploadsViewer` component is not at the same level as the `Modal` component, so we cannot control their layering. However, placing `UploadsViewer` inside the `Portal` component renders it directly in the body element, allowing me to control their z-index.

PR #840, opened 2025-03-04
✨ Handle empty search results with page loop

MontaGhanmy:


ericlinagora (approved):

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -1612,84 +1613,105 @@ export class DocumentsService {
   /**
    * Search for Drive items.
    *
-   * @param {SearchDocumentsOptions} options - the search optins.
+   * @param {SearchDocumentsOptions} options - the search options.
```
shepilov, line 1616: Could you please add e2e?
And do not reformat margins)

ericlinagora: this would be hell to e2e, you'd have to create a bunch of wrong search results, and hope they come up first

shepilov (approved):

PR #860, opened 2025-03-22
:bug: Accept username in request of API delete user

lethemanh:


tdrive/backend/node/src/core/platform/services/admin/controller/delete-user-controller.ts
```diff
@@ -13,34 +14,66 @@ export class AdminDeleteUserController {
...
   async listUsersPendingDeletion() {
-    const users = (
-      await (await this.getRepos()).user.find({}, { $gt: [["delete_process_started_epoch", 0]] })
-    ).getEntities();
-    return users.map(({ id, delete_process_started_epoch }) => [id, delete_process_started_epoch]);
+    try {
+      const users = (
+        await (await this.getRepos()).user.find({}, { $gt: [["delete_process_started_epoch", 0]] })
+      ).getEntities();
+      console.log("users: ", users);
```
ericlinagora: î

tdrive/backend/node/src/core/platform/services/admin/controller/delete-user-controller.ts
```diff
@@ -13,34 +14,66 @@ export class AdminDeleteUserController {
...
+      return users.map(({ id, delete_process_started_epoch }) => [
+        id,
+        delete_process_started_epoch,
+      ]);
+    } catch (error) {
+      console.log("error: ", error);
+      throw error;
```
ericlinagora: the whole try {} catch don't seem useful

ericlinagora (changes requested):

tdrive/backend/node/src/core/platform/services/admin/controller/delete-user-controller.ts
ericlinagora: (but if you do want to log this, there is a logger to use instead)

lethemanh: the try catch block has been removed, it used for debug in local only

tdrive/backend/node/src/core/platform/services/admin/controller/delete-user-controller.ts
```diff
@@ -13,27 +14,50 @@ export class AdminDeleteUserController {
...
+          return {
+            status: "deleting",
+            userId: existingUser.id,
+          };
+        }
       }
     } catch (err) {
-      adminLogger.error({ err, userId }, "User deletion error");
-      return "failed";
+      adminLogger.error("[DELETE USER] ", JSON.stringify({ err, userId }), "User deletion error");
```
ericlinagora: the first param given to the logger is serialised as json, and in particular the `err` key is treated specially to include the stack, so : `adminLogger.error({err, userId}, "[DELETE USER] User deletion error");`

tdrive/backend/node/src/core/platform/services/admin/controller/delete-user-controller.ts
```diff
@@ -67,4 +67,9 @@ export class AdminDeleteUserController {
     ).getEntities();
     return users.map(({ id, delete_process_started_epoch }) => [id, delete_process_started_epoch]);
   }
+
+  async getUserIdByUsername(username: string) {
+    const user = await gr.services.users.get({ username_canonical: username });
```
ericlinagora: if username isn't found, does this return a 404 ?

shepilov (approved):

shepilov (approved):

PR #868, opened 2025-04-02
Allow to share large file even it unchecked antivirus

### LINKED ISSUE #863: Be able to share file even if it isn't checked with antivirus
Right now, if the files is not checked, it can be only downloaded.

We need to have ability to share big files that are skipped with antivirus.

lethemanh:
https://github.com/user-attachments/assets/bc156650-61cf-4389-8cf7-640270d6239d

### DIFF
--- a/tdrive/backend/node/bin/config/cli.json
+++ b/tdrive/backend/node/bin/config/cli.json
@@ -128,14 +128,25 @@
     "featureManageAccess": true,
     "defaultCompany": "00000000-0000-4000-0000-000000000000",
     "defaultUserQuota": 200000000,
-    "featureAntivirus": false
+    "featureAntivirus": false,
+    "featureAvStatusAllowed": {
+      "download": ["uploaded", "safe", "scanning", "scan_failed", "malicious"],
+      "share": ["uploaded", "safe", "scanning", "skipped"],
+      "rename": ["uploaded", "safe", "scanning"],
+      "move": ["uploaded", "safe", "scanning"],
+      "rescan": ["scan_failed"],
+      "copy_link": ["uploaded", "safe", "scanning"],
+      "manage_access": ["uploaded", "safe", "scanning", "skipped"],
+      "version": ["uploaded", "safe", "scanning"]
+    }
   },
   "av": {
     "host": "av",
     "port": 3310,
     "debugMode": false,
     "timeout": 2000,
-    "maxFileSize": 26214400
+    "maxFileSize": 4294967295,
+    "deleteInfectedFiles": false
   },
   "applications": {
     "grid": [
--- a/tdrive/backend/node/src/services/av/service/index.ts
+++ b/tdrive/backend/node/src/services/av/service/index.ts
@@ -3,7 +3,7 @@ import { getLogger, logger, TdriveLogger } from "../../../core/platform/framewor
 import NodeClam from "clamscan";
 import { AVStatus, DriveFile } from "src/services/documents/entities/drive-file";
 import { FileVersion } from "src/services/documents/entities/file-version";
-import { DriveExecutionContext } from "src/services/documents/types";
+import { DriveExecutionContext, NotificationActionType } from "../../documents/types";
 import globalResolver from "../../../services/global-resolver";
 import { getFilePath } from "../../files/services";
 import { getConfigOrDefault } from "../../../utils/get-config";
@@ -14,7 +14,8 @@ export class AVServiceImpl implements TdriveServiceProvider, Initializable {
   av: NodeClam = null;
   logger: TdriveLogger = getLogger("Antivirus Service");
   avEnabled: boolean = getConfigOrDefault<boolean>("drive.featureAntivirus", false);
-  private MAX_FILE_SIZE: number = getConfigOrDefault<number>("av.maxFileSize", 26214400); // 25 MB
+  deleteInfectedFileEnabled: boolean = getConfigOrDefault<boolean>("av.deleteInfectedFiles", false);
+  private MAX_FILE_SIZE: number = getConfigOrDefault<number>("av.maxFileSize", 4294967295); // 4GB
 
   async init(): Promise<this> {
     try {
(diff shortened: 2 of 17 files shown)


tdrive/backend/node/src/services/av/service/index.ts
```diff
@@ -61,7 +61,7 @@ export class AVServiceImpl implements TdriveServiceProvider, Initializable {
         this.logger.info(
           `File ${file.id} is too large (${file.upload_data.size} bytes) to be scanned. Skipping...`,
         );
-        return "skipped";
+        return "uploaded_large_file";
```
ericlinagora: all skipped are because they are too large, what's the distinction you're making here ?

lethemanh: You're right, firstly I'm not sure if the `skipped` status is used for large files so I created a new one. I've removed it and used `skipped`

ericlinagora (approved):

tdrive/backend/node/bin/config/cli.json
```diff
@@ -128,14 +128,25 @@
...
+      "version": ["uploaded", "safe", "scanning"]
+    }
   },
   "av": {
     "host": "av",
     "port": 3310,
     "debugMode": false,
     "timeout": 2000,
-    "maxFileSize": 26214400
+    "maxFileSize": 4294967296,
```
ericlinagora: this is UINT32_MAX+1, just to be on the safe side with 32bit all through the clamav chain and the dockers and fs etc, please change it to just 1 less, 4294967295, so it fits in a uint32. (same in the other json config)

tdrive/backend/node/src/services/av/service/index.ts
```diff
@@ -14,7 +14,8 @@ export class AVServiceImpl implements TdriveServiceProvider, Initializable {
   av: NodeClam = null;
   logger: TdriveLogger = getLogger("Antivirus Service");
   avEnabled: boolean = getConfigOrDefault<boolean>("drive.featureAntivirus", false);
-  private MAX_FILE_SIZE: number = getConfigOrDefault<number>("av.maxFileSize", 26214400); // 25 MB
+  deleteInfectedFileEnabled: boolean = getConfigOrDefault<boolean>("av.deleteInfectedFiles", false);
+  private MAX_FILE_SIZE: number = getConfigOrDefault<number>("av.maxFileSize", 4294967296); // 4GB
```
ericlinagora: uint32_max+1

tdrive/backend/node/src/services/av/service/index.ts
```diff
@@ -95,4 +103,27 @@ export class AVServiceImpl implements TdriveServiceProvider, Initializable {
...
+        context,
+        item: {
+          ...item,
+          company_id: context.company.id,
+        } as DriveFile,
+        type: NotificationActionType.DIRECT,
+        notificationReceiver: context.user.id,
+        notificationEmitter: "",
+      });
+      this.logger.info(`Sent notification to ${context.user.id} about file deletion`);
```
ericlinagora, line 123: if you use the first param as an object for the logger, in graphana later it makes it easy to filter by those fields, and they are displayed separately from the message. eg.

```ts
this.logger.info({user_id: context.user.id, notification_type: 'av_caused_file_deletion'}, `Sent notification to ${context.user.id} about file deletion`);
```

ericlinagora: (not important here, just, a tool for you to know about)

tdrive/backend/node/src/services/documents/services/index.ts
```diff
@@ -1833,7 +1838,10 @@ export class DocumentsService {
     item.av_status = status;
     await this.repository.save(item);
 
-    if (["malicious", "scan_failed"].includes(status)) {
+    if (
+      (!globalResolver.services.av?.deleteInfectedFileEnabled && status === "malicious") ||
```
ericlinagora, line 1842: so we only notify if it's infected and we're deleting, or in all cases if the scan itself had a technical failure? is that the demand from product ? seems odd

tdrive/backend/node/bin/config/cli.json
```diff
@@ -128,14 +128,25 @@
     "featureManageAccess": true,
     "defaultCompany": "00000000-0000-4000-0000-000000000000",
     "defaultUserQuota": 200000000,
-    "featureAntivirus": false
+    "featureAntivirus": false,
+    "featureAvStatusAllowed": {
```
ericlinagora, line 132: why no custom-environment-variables entry for this ? there's a format for json values, eg: https://github.com/linagora/twake-drive/blob/f0b6315d762f6bf954601c90713a7f321f9a582b/tdrive/backend/node/config/custom-environment-variables.json#L92-L95

ericlinagora (changes requested):
Did you get approval to skip the ru and vi locales ?

tdrive/backend/node/src/services/documents/services/index.ts
shepilov: @lethemanh yes, we should notify when the file as deleted, and also when we keep file as well, but just text of notification will be different

lethemanh: @shepilov @ericlinagora we still notify for user about the deleted file but it trigger when the deletion is complete, as you can see here https://github.com/linagora/twake-drive/pull/868/files#diff-16ab5e2156265073c1d9b67433b2f7e1bae90fa6953482f03d26258401dbc7e2R114

tdrive/backend/node/src/services/av/service/index.ts
```diff
@@ -56,7 +57,9 @@ export class AVServiceImpl implements TdriveServiceProvider, Initializable {
         this.logger.error(`File ${version.file_metadata.external_id} not found`);
         throw AVException.fileNotFound(`File ${version.file_metadata.external_id} not found`);
       }
+
       // check if the file is too large
+      logger.debug("MAX_FILE_SIZE: ", this.MAX_FILE_SIZE);
```
ericlinagora: is this needed ? I mean it could be a good value to have in a logger's first param if there's a decision to skip or an error, but for every file to log just that doesn't seem valuable in prod

ericlinagora (changes requested):

tdrive/backend/node/src/services/av/service/index.ts
lethemanh: I will remove it, just use for debugging before

ericlinagora (approved):

PR #870, opened 2025-04-04
✨ Support tool, zips folder and puts it to users home directory

shepilov:


tdrive/backend/node/src/cli/cmds/dev_cmds/zip_folder.ts
```diff
@@ -0,0 +1,138 @@
...
+  return user;
+}
+
+async function zipFolder(
+  toZip: DriveItemDetails,
+  outputPath: string,
+  context: CompanyExecutionContext,
+) {
+  return new Promise<void>(async (resolve, reject) => {
+    return await globalResolver.services.documents.documents.download(
```
ericlinagora, line 130: The issue with that function is that it had to be rewritten to start writing immediately, time to first byte was causing downloads of large numbers to timeout (at nginx level I think). Because of this, it doesn't really detect errors like missing chunks. It does skip files if they don't actually exist in S3, silently, but any other error depends on where the zip stream and file stream are in relation to each other. I also don't think it creates a folder if it's empty, the prefix seems to be only on files - but didn't test this.

ericlinagora (approved):

tdrive/backend/node/src/cli/cmds/dev_cmds/zip_folder.ts
shepilov: It's a console tool, so timeout is not a problem in this case, and in case of missing files, we will have logs

ericlinagora: No, but the fix to the timeout issue, means that you can get out of that function with an incomplete zip file, incomplete files inside the zip, and missing empty folders. For the logs we still have a bunch of `catch (e) { logger.debug/info("some error"); return; }`

PR #873, opened 2025-04-09
:lipstick: Rework anti virus UI

### LINKED ISSUE #821: [UI] rework antivirus UI
- In browser rows, use text hints to show difference between skipped and scan failed, and clarify malicious
- Download alert needs:
    - More alarming presentation for the presence of known malicious, to distinguish from
    - A presentation more like actual, explaining that the scan was skipped, didn't run yet or failed
    - More obvious buttons: "Move to Trash" and "Download anyway" in a popping colour (+ currently cancel label isn't translated)

lethemanh:
CleanShot 2025-04-09 at 14 55 40@2x
CleanShot 2025-04-09 at 14 55 53@2x
CleanShot 2025-04-09 at 14 56 07@2x
CleanShot 2025-04-09 at 14 57 19@2x


tdrive/frontend/public/locales/en.json
```diff
@@ -35,6 +35,8 @@
     "components.add_mails_workspace.text_area_placeholder": "Enter emails of your users*",
     "components.alert.confirm": "Confirm your action",
     "components.alert.confirm_click": "Confirm your action by clicking on OK.",
+    "components.alert.ok_text": "OK",
+    "components.alert.cancel_text": "Cancel",
```
ericlinagora: there's already a generic one for cancel: `general.cancel`, maybe that's where `ok` should go too instead ? I'm surprised we don't already have one

tdrive/frontend/public/locales/en.json
```diff
@@ -224,6 +227,10 @@
     "scenes.app.drive.document_row.av_malicious": "Malicious",
     "scenes.app.drive.document_row.av_skipped": "Skipped",
     "scenes.app.drive.document_row.av_scan_failed": "Scan failed",
+    "scenes.app.drive.document_row.av_scanning_hint": "This item is scanning",
+    "scenes.app.drive.document_row.av_malicious_hint": "This item is malicious",
```
ericlinagora: Can we say "potentially contains malware/a virus" or something like that ? Malicious seems a bit of a technical catch all jargon most people wouldn't get. Same in french, malveillant doesnt really work =/ The exact wording in french might be something Anton needs to check with Patrick

PR #910, opened 2025-10-15
🐛 fix(files migration): handle big files and users with countless files

rezk2ll:


tdrive/backend/node/src/cli/cmds/migration_cmds/migrate-files.ts
```diff
@@ -62,7 +63,17 @@ const purgeIndexesCommand: yargs.CommandModule<unknown, unknown> = {
...
           const userCompany = DEFAULT_COMPANY;
-          const userFiles = await documentsRepo.find({ creator: user.id, is_directory: false });
+          const userFiles = await documentsRepo.find(
+            {
+              creator: user.id,
+              is_directory: false,
+              migrated: false,
+            },
+            {
+              pagination: { limitStr: "10000" },
```
Crash--, line 74: add a big TODO here. Maybe for customer we will have to handle pagination.

Crash-- (approved):

